PR #3, opened 2018-07-13
readme: add more use-cases

sirupsen:
Some of my favorite use-cases other than waiting for lag :) Curious about your future vision for the open-source suite? Are you planning to release e.g. database throttling as a plugin to this stack? 👂 

Super psyched this is finally open source :blobheart:

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ The job would run fairly quickly when you only have a hundred User records. But 
 
 With frequent deploys and worker restarts, it would mean that a job will be either lost of started from the beginning. Some records (especially those in the beginning of the relation) will be processed more than once.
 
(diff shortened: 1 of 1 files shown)


kirs:
>Are you planning to release e.g. database throttling as a plugin to this stack?

This would be awesome for Iteration as a product, but on another hand the one we use at Shopify is also very specific to us:

* it assumes you have more than one connection (at least reader and writer)
* it assumes that you're using MySQL
* it assumes that you're using `pt-heartbeat` to report replication lag

I'm looking forward to ship the gem with things like throttling, but it could also be ambigious to attempt implementing more throttling strategies than MySQL and `pt-heartbeat`.

I wonder if throttling could be so abstract that it allowed arbitrary configuration and custom checks. But first I'd like to hear thoughts of the early adopters.

PR #11, opened 2018-08-29
Fix typo and tweak language

chrisalbrecht:
👋 

Just noticed this typo (of => or) while reading through the docs. I also thought that its a bit clearer to say the job will be restarted from the beginning.


kirs:
Thanks!

PR #18, opened 2018-10-29
Do not mark as complete when throttled

### LINKED ISSUE #17: on_complete callbacks are called when ThrottlingEnumerator delays the job
If any enumerator is wrapped into ThrottlingEnumerator that aborts the iteration early, Iteration API will consider it as "complete" and would call `on_complete` callbacks. As a result `on_complete` callbacks will be called more than once.

GustavoCaso:
Fixes #17 

Make sure that `on_complete` is only called with the job has been completed.

### DIFF
--- a/test/unit/active_job_iteration_test.rb
+++ b/test/unit/active_job_iteration_test.rb
@@ -351,7 +351,7 @@ module JobIteration
       assert cursor
 
       assert_equal 0, ActiveRecordIterationJob.on_complete_called
-      assert_equal 4, ActiveRecordIterationJob.on_shutdown_called
+      assert_equal 2, ActiveRecordIterationJob.on_shutdown_called
     end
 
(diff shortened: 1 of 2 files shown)


test/unit/active_job_iteration_test.rb
```diff
@@ -351,7 +351,7 @@ def test_active_record_job
       assert cursor
 
       assert_equal 0, ActiveRecordIterationJob.on_complete_called
-      assert_equal 2, ActiveRecordIterationJob.on_shutdown_called
+      assert_equal 4, ActiveRecordIterationJob.on_shutdown_called
```
kirs: Any idea why this changed from 2 to 4?

GustavoCaso: Good question. Let me check. because based on the code it shouldn't have changed the value

PR #21, opened 2018-11-06
Only skip complete callback if marked by the user

GustavoCaso:
Previously we skipped the complete callback if something stops the iteration of the job using `throw` and `catch`, but we need to make the execution of `complete` callback by default, but we also need to provide a way to skip them.

This PR leverage that by using `throw(:abort, :skip_complete_callbacks)`

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -72,6 +72,11 @@ module JobIteration
       interruptible_perform(*params)
     end
 
+    def retry_job(*)
+      @retried = true
+      super
+    end
+
     private
 
     def enumerator_builder
@@ -108,7 +113,7 @@ module JobIteration
 
       run_callbacks :shutdown
 
-      if completed
+      if run_complete_callbacks?(completed)
         run_callbacks :complete
         output_interrupt_summary
       end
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -107,6 +107,35 @@ end
...
+  def each_iteration(record)
+    self.class.records_performed << record
+    throw(:abort) if self.class.records_performed.size == 2
+  end
+end
+```
+
+If you want to terminate the execution of the job and skip the complete callback use `throw(:abort, :skip_complete_callbacks)`
+
+Using `throw` and `catch` instead of Exceptions is a deliberated decision, by design using exceptions as control flow is a bad design pattern, also the performance implications of using exceptions on a library like this one are important. Every time we raise an Exception ruby has to build a stack trace, so avoiding it will help us.
```
kirs: I'm not sure if we need this clarification in the README. It would become quite large if we documented every design decision there.

kirs: Since it's not a great design and we're not sure if we'll have those callbacks in future releases of gem (https://github.com/Shopify/job-iteration/issues/19), I'd avoid promoting it to users.

GustavoCaso: Fantastic. Let's remove it.

PR #32, opened 2019-04-15
use enqueue instead of retry_job to reenqueue jobs

GustavoCaso:
@kirs I wanted to see if we can remove the call for `retry_job` and all the test will pass :)

I wanted to remove the use of `retry_job` because I plan on overwrite it on core to increase the number of attempts from the `Retry` module, to make sure are not retrying more than they should.


lib/job-iteration/iteration.rb
```diff
@@ -150,7 +145,9 @@ def reenqueue_iteration_job
       self.times_interrupted += 1
 
       self.already_in_queue = true if respond_to?(:already_in_queue=)
-      retry_job
+
+      enqueue unless @enqueued
```
kirs: You no-op enqueueing if `@enqueued` is set, but what about the rest of side-effects in that method? Why do they need to run if the job was already enqueued?

rafaelfranca: Unless I'm missing something you need the `defined?(@enqueued)` otherwise you will get a Ruby warning.

rafaelfranca:
I'm not against this change, if the goal if to make `retry_job` make sure we don't retry more than we should and this is the only way to do 👍, but maybe we need a `re_enqueue_job` method as well to make this distinction clea.

PR #35, opened 2019-06-07
Make sure `cursor` is a keyword argument

GustavoCaso:
We found cases of people using `cursor` argument, not as a keyword argument.

This can lead to unexpected behaviour.

[Discourse ticket](https://discourse.shopify.io/t/maintenancetask-throttled-restarts-from-the-beginning/3490)

This change introduces an extra level of checking so when running the
job, we make sure the signature of build_enumerator has the `cursor` keyword
argument

@Shopify/job-patterns

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -176,7 +176,13 @@ module JobIteration
         )
       end
 
-      unless respond_to?(:build_enumerator, true)
+      if respond_to?(:build_enumerator, true)
+        parameters = method(:build_enumerator).parameters
+        unless valid_cursor_parameter?(parameters)
+          raise ArgumentError, "Iteration job (#{self.class}) #build_enumerator " \
+          "expects the keyword argument `cursor`"
+        end
+      else
         raise ArgumentError, "Iteration job (#{self.class}) must implement #build_enumerator " \
           "to provide a collection to iterate"
       end
@@ -213,6 +219,18 @@ module JobIteration
       # be executed
       when false, :skip_complete_callbacks then false
       end
+    end
+
+    def valid_cursor_parameter?(parameters)
+      # this condition is when people use the splat operator.
+      # def build_enumerator(*)
+      return true if parameters == [[:rest]]
+
+      parameters.each do |parameter_type, parameter_name|
+        next unless parameter_name == :cursor
+        return true if parameter_type == :keyreq
+      end
+      false
     end
   end
 end
--- a/test/unit/iteration_test.rb
+++ b/test/unit/iteration_test.rb
@@ -3,7 +3,78 @@
 require "test_helper"
 
 class JobIterationTest < IterationUnitTest
+  class JobWithNoMethods < ActiveJob::Base
+    include JobIteration::Iteration
+  end
+
+  class JobWithRightMethods < ActiveJob::Base
+    include JobIteration::Iteration
+    def build_enumerator(_params, cursor:)
+      enumerator_builder.build_times_enumerator(2, cursor: cursor)
+    end
+
+    def each_iteration(*)
+    end
+  end
+
+  class JobWithRightMethodsButMissingCursorKeywordArgument < ActiveJob::Base
+    include JobIteration::Iteration
+    def build_enumerator(params, cursor)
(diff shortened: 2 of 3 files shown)


lib/job-iteration/iteration.rb
```diff
@@ -214,5 +220,18 @@ def run_complete_callbacks?(completed)
       when false, :skip_complete_callbacks then false
       end
     end
+
+    def valid_cursor_parameter?(parameters)
+      # this condition is when people use the splat operator.
+      # def build_enumerator(*)
+      return true if parameters == [[:rest]]
```
BoGs: TIL

lib/job-iteration/iteration.rb
```diff
@@ -214,5 +220,18 @@ def run_complete_callbacks?(completed)
...
+
```
djmortonShopify: Wouldn't this be slightly simpler?

```
    parameters.each do |parameter_type, parameter_name|
      next unless parameter_name == :cursor
      return true if parameter_type == :keyreq
    end
    false
  end
```

test/unit/iteration_test.rb
```diff
@@ -3,7 +3,78 @@
...
+    push(JobWithRightMethods, 'walrus' => 'best')
+    work_one_job
+  end
+
+  def test_jobs_that_pass_splat_argument_to_build_enumerator_will_not_raise
+    push(JobWithRightMethodsUsingSplatInTheArguments, {})
+    work_one_job
+  end
+
+  def test_jobs_that_do_not_define_build_enumerator_or_each_iteration_will_not_raise
```
djmortonShopify: `each_iteration_will_not_raise` but the test has an `assert_raises(ArgumentError)`

lib/job-iteration/iteration.rb
```diff
@@ -176,7 +176,13 @@ def assert_implements_methods!
         )
       end
 
-      unless respond_to?(:build_enumerator, true)
+      if respond_to?(:build_enumerator, true)
+        parameters = method(:build_enumerator).parameters
```
kirs: This is not strictly related to your PR or needs addressing, but it's somewhat pointless to check for arguments and the method signature on _every_ `perform` of a job. Once a job class has been validated, we know it implements methods right. This makes me believe that we should be skipping `assert_implements_methods!` for classes that has already been validated.

lib/job-iteration/iteration.rb
```diff
@@ -176,7 +176,13 @@ def assert_implements_methods!
...
+        unless valid_cursor_parameter?(parameters)
+          raise ArgumentError, "Iteration job (#{self.class}) #build_enumerator " \
+          "expects the keyword argument `cursor` please make sure that this condition is met"
```
kirs: Nitpick: `please make sure that this condition is met` may not the language that's typically used in exception messages. "expects the keyword argument `cursor`" is probably enough for developers to understand that's a condition that they have to satisfy :)

kirs:
Looks like `method(:build_enumerator).parameters` doesn't carry much performance overhead:

```
require 'benchmark/ips'

def build_enumerator
end

Benchmark.ips do |x|
  x.config(:time => 3, :warmup => 1)

  x.report("respond_to") do
    respond_to?(:build_enumerator)
  end
  x.report("respond_to + parameters") do
    respond_to?(:build_enumerator) && method(:build_enumerator).parameters
  end

  x.compare!
end
```

```
Warming up --------------------------------------
          respond_to   270.254k i/100ms
respond_to + parameters
                       264.253k i/100ms
Calculating -------------------------------------
          respond_to      8.025M (± 2.1%) i/s -     24.323M in   3.032362s
respond_to + parameters
                          8.136M (± 3.2%) i/s -     24.576M in   3.024205s

Comparison:
respond_to + parameters:  8135920.7 i/s
          respond_to:  8024793.9 i/s - same-ish: difference falls within error
```

lib/job-iteration/iteration.rb
GustavoCaso: That is a great point, we could have a local cache of checked classes and avoid having to call `assert_implements_methods!` unnecessary. I will explore this on another PR.

PR #38, opened 2019-07-24
Fix count_rows_in_file for ruby 2.6.3

peterzhu2118:
[`CSV#path`](https://ruby-doc.org/stdlib-2.6.3/libdoc/csv/rdoc/CSV.html#method-i-path) was added to Ruby 2.6.3 and returns `nil` when the path of the CSV does not exist (instead of raising a `NoMethodError`). Ran into an issue where this method was returning `-1` in 2.6.3 when the path was `nil`.

Also bumps the patch version.

### DIFF
--- a/lib/job-iteration/csv_enumerator.rb
+++ b/lib/job-iteration/csv_enumerator.rb
@@ -49,6 +49,7 @@ module JobIteration
     private
 
     def count_rows_in_file
+      # TODO: Remove rescue for NoMethodError when Ruby 2.6 is no longer supported.
       begin
         filepath = @csv.path
       rescue NoMethodError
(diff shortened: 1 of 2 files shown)


lib/job-iteration/csv_enumerator.rb
```diff
@@ -55,6 +55,9 @@ def count_rows_in_file
         return
       end
 
+      # Behaviour of CSV#path changed in Ruby 2.6.3 (returns nil instead of raising NoMethodError)
+      return if filepath.nil?
```
rafaelfranca: Can we add a smart todo note to remove the `rescue`?

rafaelfranca:
```suggestion
      return unless filepath
```

Issue #37, opened 2019-07-18
Do not validate enumerator arguments every time we perform a job


GustavoCaso:
Every time we enqueue a job we validate arguments using `assert_implements_methods!`

Is not a very expensive operation, but once we enqueue the first job we know that the rest of the jobs of that class are valid.

We need to find a way to skip the method signature validations once we have validated it.

PR #39, opened 2019-07-30
Refactor the class method validation logic

anuragshopify:
This is in regards to - https://github.com/Shopify/job-iteration/issues/37

This PR moves the method signature checks (i.e `assert_implements_methods!`) to the job initialization step.

This method is really looking at definition level stuff i.e whether the
job class has the necessary methods and the method declarations are
correct. As such, I think having this sort of check in the constructor
makes more sense.
This also updates all the tests accordingly.

If this isn't the case/there's a case for not having it in the
constructor we can just use a boolean to not re-check the signatures on
every invocation of `perform`.

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -47,6 +47,7 @@ module JobIteration
       super
       self.times_interrupted = 0
       self.total_time = 0.0
+      assert_implements_methods!
     end
 
     def serialize # @private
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -18,6 +18,7 @@ gem 'redis'
 gem 'database_cleaner'
 
 gem 'pry'
+gem 'pry-nav'
```
GustavoCaso: Do we need this?

GustavoCaso:
Thanks for working on this ❤️ 

I left a message for you to review.

Gemfile
anuragshopify: It helped with navigating the debugger. I can remove it if you like.

GustavoCaso: Yeah, let's remove it. Please ❤️

PR #45, opened 2019-12-11
Throttle enumerator

kirs:
Let's finally extract throttle enumerator from shopify/shopify. It might be useful for other apps that are building something that's supposed to throttle on DB health. I'm specifically thinking of https://github.com/Shopify/shopify-cloud/pull/155 here.

### DIFF
--- a/guides/throttling.md
+++ b/guides/throttling.md
@@ -1,0 +1,46 @@
+Iteration comes with a special wrapper enumerator that allows you to throttle iterations based on external signal (e.g. database health).
+
+Consider this example:
+
+```ruby
+class InactiveAccountDeleteJob < ActiveJob::Base
+  include JobIteration::Iteration
+
+  def build_enumerator(_params, cursor:)
+    enumerator_builder.active_record_on_batches(
+      Account.inactive,
+      cursor: cursor
+    )
+  end
+
+  def each_iteration(batch, _params)
+    Account.where(id: batch.map(&:id)).delete_all
+  end
+end
+```
+
+For an app that keeps track of customer accounts, it's typical to purge old data that's no longer relevant for storage.
+
+At the same time, if you've got a lot of DB writes to perform, this can cause extra load on the database and slow down other parts of your service.
+
+You can change `build_enumerator` to wrap enumeration on DB rows into a throttle enumerator, which takes signal as a proc and enqueues the job for later in case the proc returned `true`.
+
+```ruby
+def build_enumerator(_params, cursor:)
+  enumerator_builder.build_throttle_enumerator(
+    enumerator_builder.active_record_on_batches(
+      Account.inactive,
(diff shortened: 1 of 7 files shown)


guides/throttling.md
```diff
@@ -0,0 +1,46 @@
...
+  )
+end
+```
+
+Note that it's up to you to implement `DatabaseStatus.unhealthy?` that works for your database choice. At Shopify, a helper like `DatabaseStatus` checks the following MySQL metrics:
+
+* Replication lag across all regions
+* DB threads
+* DB is available for writes (otherwise indicates a failover happening)
+* [Semian](https://github.com/shopify/semian) curciuts
```
moechaieb: `open circuits`

lib/job-iteration/throttle_enumerator.rb
```diff
@@ -0,0 +1,46 @@
...
+            throw(:abort, :skip_complete_callbacks)
+          end
+
+          yielder.yield(*val)
+        end
+      end
+    end
+
+    def should_throttle?
+      @throttle_on.call
```
moechaieb: Is it worth being defensive around this being a proc? What about returning a boolean?

kirs: Do you mean validating the return of a proc? Or validating that `@throttle_on` is a proc?

GustavoCaso: @moechaieb are. you saying not passing a proc as the `throttle_on` variable?

GustavoCaso:
I can take care of deploying it and updating core with the latest version.

lib/job-iteration/throttle_enumerator.rb
moechaieb: > Do you mean validating the return of a proc? Or validating that @throttle_on is a proc?

I'm asking if it's worth doing both. I'm usually not in favour of overly defensive programming. I think you know best what to do here, and my comment was just a suggestion.

kirs: It's not always a proc, it can be anything that responds to `call` like:

```ruby
"".method(:upcase)
=> #<Method: String#upcase>
```

We could make it a bit more friendly and raise if the thing doesn't respond to `call`, but I find it pretty unlikely scenario to optimize for.

GustavoCaso: The problem with passing something that is precomputed is that it would be the same value for all the iteration cycle of that job.

kirs: For a string in my example, yes but for `Database::Status.method(:healthy?)` it kinda doesn't matter.

More over, `throttle_on: Database::Status.method(:healthy?)` vs `throttle_on: -> { Database::Status.healthy? }` saves you an extra stack.

PR #49, opened 2020-03-18
Log when enumerator has nothing to iterate

djmortonShopify:
If an Enumerator is built and has nothing to iterate, it can be confusing for developers.  This is particularly true with ActiveRecord enumerators, where developers can be tripped up by default scopes excluding records they don't realize will be excluded.  This PR adds a log message if we attempt to iterate an enumerator with no records to hopefully hint to developers that their enumerator did not actually return anything to help them pinpoint the source of their problem.

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -116,8 +116,10 @@ module JobIteration
 
     def iterate_with_enumerator(enumerator, arguments)
       arguments = arguments.dup.freeze
+      found_record = false
       enumerator.each do |object_from_enumerator, index|
         record_unit_of_work do
+          found_record = true
           each_iteration(object_from_enumerator, *arguments)
           self.cursor_position = index
         end
@@ -127,6 +129,11 @@ module JobIteration
         reenqueue_iteration_job
         return false
       end
+
+      logger.info(
+        "[JobIteration::Iteration] Enumerator found nothing to iterate! " \
+        "(times_interrupted: #{times_interrupted}, cursor_position: #{cursor_position})"
+      ) unless found_record
 
       true
     end
--- a/test/unit/active_job_iteration_test.rb
+++ b/test/unit/active_job_iteration_test.rb
@@ -585,6 +585,15 @@ module JobIteration
       end
     end
 
+    def test_log_no_iterations
+      Product.delete_all
+      push(ActiveRecordIterationJob)
+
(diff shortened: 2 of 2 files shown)


test/unit/active_job_iteration_test.rb
```diff
@@ -585,6 +585,16 @@ def test_log_completion_data
...
     end
 
+    def test_log_no_iterations
+      Product.delete_all
+      push(ActiveRecordIterationJob)
+
+      assert_logged(%([JobIteration::Iteration] Enumerator found nothing to iterate!)) do
+        work_one_job
+      end
+
```
moechaieb: ✂️

moechaieb:
Thoughts on moving this one level up the stack and adding this logging to Iteration API so that we log when any enumerator is empty?

GustavoCaso:
> Thoughts on moving this one level up the stack and adding this logging to Iteration API so that we log when any enumerator is empty?
@moechaieb this as low as you can get, right? What do you mean with the Iteration API?

djmortonShopify:
The ActiveRecord case is the one that seems to come up most often, but I'll look at moving it up and see if it makes sense.

djmortonShopify:
@moechaieb Sorry... my brain blanked there... My test is for activerecord enumerators, but the actual code is in `iteration.rb` and should handle all enumerators.

GustavoCaso:
@moechaieb @djmortonShopify this change deal with all enumerators. 

We first `build_enumerator` which is agnostic of the enumerator and then we iterate over it.

lib/job-iteration/iteration.rb
```diff
@@ -128,6 +130,11 @@ def iterate_with_enumerator(enumerator, arguments)
         return false
       end
 
+      logger.info(
+        "[JobIteration::Iteration] Enumerator found nothing to iterate! " \
+        "(times_interrupted: #{times_interrupted}, cursor_position: #{cursor_position})"
```
kirs: Why not use tagged logging like we do everywhere? E.g. times_interrupted=#{times_interrupted} cursor_position=#{cursor_position}`?

kirs:
The version you guys pushed should solve the immediate problem. An alternative way that I thought of would be to count number of iterations that has been done in a run and _always_ log that number, which would make the log output consistent (you'd see it with all jobs) and maybe help to debug some other problems in the future.

PR #52, opened 2020-05-13
Fix off-by-one-error in CsvEnumerator

sunyoung-shopify:
### What are you trying to accomplish?
I'm addressing an off-by-one-error in the iteration of the `CsvEnumerator`.  I noticed the [bug](https://discourse.shopify.io/t/iteration-api-cursor-points-to-same-record-despite-seemingly-completing-its-iteration/8229/3) when I was using `JobIteration::CsvEnumerator.new(csv).rows(cursor: cursor)` in a job that imports records. When the job would resume after interruption, it would begin by re-processing the record that was *just* sucessfully imported prior to interruption. That is, the job would re-process the last iteration instead of the moving onto the *next* iteration as expected. This problem is happening because the `CsvEnumerator` passes the given `cursor` position directly into `#drop` without incrementing it first. `#drop` expects the number of elements to remove from an array, not the index of the element we want to drop. Since we just pass in the index of the last sucessful iteration and `#drop` isn't positional, we end up removing all the elements *before* the last iteration, but not the last iteration itself.

I noticed the `batches` version of the `CsvEnumerator` had the same problem, so I went ahead and updated that method and its tests too.

A quick shoutout to @alanly for working with me on this and spotting the error in the code 🙏 

### What approach did you choose and why?
- I handle the `cursor` in the same way as the [`#build_array_enumerator`](https://github.com/Shopify/job-iteration/blob/master/lib/job-iteration/enumerator_builder.rb#L55-L60). I increment the cursor by 1 before passing it to `#drop`, unless the cursor is `nil`, in which case `0` is passed instead
- I updated the existing tests to now work with the fix
- I removed the `considers cursor: nil as the start` tests because they're now redundant. We implicitly test that the enumerator properly handles `nil` in the `yields every record/batch with their cursor position` tests

### What should reviewers focus on?
Are there any other tests or files that should be updated?

### DIFF
--- a/lib/job-iteration/csv_enumerator.rb
+++ b/lib/job-iteration/csv_enumerator.rb
@@ -30,12 +30,10 @@ module JobIteration
     # Constructs a enumerator on CSV rows
     # @return [Enumerator] Enumerator instance
     def rows(cursor:)
-      drop = cursor.nil? ? 0 : cursor + 1
-
       @csv.lazy
         .each_with_index
-        .drop(drop)
-        .to_enum { count_rows_in_file }
+        .drop(count_of_processed_rows(cursor))
+        .to_enum { count_of_rows_in_file }
     end
 
     # Constructs a enumerator on batches of CSV rows
@@ -44,13 +42,13 @@ module JobIteration
       @csv.lazy
         .each_slice(batch_size)
         .each_with_index
-        .drop(cursor.to_i)
-        .to_enum { (count_rows_in_file.to_f / batch_size).ceil }
+        .drop(count_of_processed_rows(cursor))
+        .to_enum { (count_of_rows_in_file.to_f / batch_size).ceil }
     end
 
     private
 
-    def count_rows_in_file
+    def count_of_rows_in_file
       # TODO: Remove rescue for NoMethodError when Ruby 2.6 is no longer supported.
       begin
         filepath = @csv.path
@@ -64,6 +62,10 @@ module JobIteration
       count = %x(wc -l < #{filepath}).strip.to_i
       count -= 1 if @csv.headers
       count
+    end
+
+    def count_of_processed_rows(cursor)
+      cursor.nil? ? 0 : cursor + 1
     end
   end
 end
--- a/lib/job-iteration/csv_enumerator.rb
+++ b/lib/job-iteration/csv_enumerator.rb
@@ -30,9 +30,11 @@ module JobIteration
     # Constructs a enumerator on CSV rows
     # @return [Enumerator] Enumerator instance
     def rows(cursor:)
+      drop = cursor.nil? ? 0 : cursor + 1
+
       @csv.lazy
         .each_with_index
-        .drop(cursor.to_i)
+        .drop(drop)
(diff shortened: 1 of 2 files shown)


djmortonShopify:
Is there any way the off-by-one nature of this is related to the fact that often CSV files will have a first row containing headers, rather than data? In that way, it might be different that just mimicking what the array enumerator does. Was this tested with both CSV files that contain a header row and CSV files that don't?

alanly:
>_Is there any way the off-by-one nature of this is related to the fact that often CSV files will have a first row containing headers, rather than data?_

I believe the behaviour around the header row is dictated by the CSV instance that's passed to the enumerator constructor.

There seems to be existing tests that asserts the behaviour when headers are parsed and when they're not parsed:

https://github.com/Shopify/job-iteration/blob/c050da93116b3301c936c70b24df37fb15902099/test/unit/csv_enumerator_test.rb#L44-L48

https://github.com/Shopify/job-iteration/blob/c050da93116b3301c936c70b24df37fb15902099/test/unit/csv_enumerator_test.rb#L60-L63

lib/job-iteration/csv_enumerator.rb
```diff
@@ -30,19 +30,23 @@ def initialize(csv)
     # Constructs a enumerator on CSV rows
     # @return [Enumerator] Enumerator instance
     def rows(cursor:)
```
alanly, line 32: Is it worth documenting what `cursor` means for the iteration on resume (specifically that it represents the last _successfully_ processed row?)

GustavoCaso:
I tested locally inside an `irb` session, and I think it makes total sense the change 😄 

```ruby
csv = [['a', 'a', 'a'], ['bb', 'bb', 'bb'], ['c']]

# Lets start the iteration, cursor is nil
# nil.to_i gets converted to 0

enum = csv.each_with_index.drop(0).to_enum
=> #<Enumerator: [[["a", "a", "a"], 0], [["bb", "bb", "bb"], 1], [["c"], 2]]:each>

# We iterate and the job gets halted after processing element 1
# So now cursor is 1

# If we fetch 1 as it is it will return the previously processed element 
enum = csv.each_with_index.drop(1).to_enum
=> #<Enumerator: [[["bb", "bb", "bb"], 1], [["c"], 2]]:each>

# But if we apply the change from the PR returns the right elements 
enum = csv_dup.each_with_index.drop(2).to_enum
=> #<Enumerator: [[["c"], 2]]:each>
```

Great work 🎉

lib/job-iteration/csv_enumerator.rb
```diff
@@ -30,19 +30,23 @@ def initialize(csv)
...
+      drop = cursor.nil? ? 0 : cursor + 1
```
GustavoCaso, line 33: Could we extract this to a method that expresses the intent of this logic? 

`advance_cursor`?
`move_cursor`?

I'm not very good with names if I'm honest 😄

lib/job-iteration/csv_enumerator.rb
GustavoCaso: Yeah it would make sense

PR #59, opened 2020-11-13
prototype wrapping of query execution for AR cursor

eljojo:
hey @kirs @byroot i'm doing a prototype here, no real intentions of merging, any comments welcome. I am doing this for a personal project.

here's the context: I have a vanilla rails app and I'm trying to use multiple connections within the same job. I am using rails' [connected_to](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionHandling.html#method-i-connected_to) block mechanism.

I am running into a problem where I switch connections during each_iteration, which makes the job finish after the first batch is done. It seems that AR switches the default to connection to whatever `connected_to` block was called last.

In any case, this is an attempt to make it possible to pass a block around AR enumerator builder, so I can wrap the query execution in the appropriate connected_to block.

### DIFF
--- a/lib/job-iteration/active_record_cursor.rb
+++ b/lib/job-iteration/active_record_cursor.rb
@@ -18,7 +18,7 @@ module JobIteration
       end
     end
 
-    def initialize(relation, columns = nil, position = nil)
+    def initialize(relation, columns = nil, position = nil, around_query = nil)
       columns ||= "#{relation.table_name}.#{relation.primary_key}"
       @columns = Array.wrap(columns)
       self.position = Array.wrap(position)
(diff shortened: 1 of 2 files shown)


kirs:
Hmm. I'd love to make it work nicely with AR's multi-database feature. Ideally with developers not having to pass `around_query:`.

I think we could scope this out to vanilla ActiveRecord gist like:

```ruby
enum = Enumerator.new do |yielder|
  ActiveRecord::Base.connected_to(role: :foo) do
     Shop.find_each { |shop| yielder.yield(record) }
  end
end

enum.each do
  ActiveRecord::Base.connected_to(role: :bar) do
    # do something else
  end
end
```

Would that work? If not maybe it's an edge case of ActiveRecord that we could take care of upstream.

PR #60, opened 2020-12-09
Add warning about custom enumerator yields

grcooper:
This ended up causing a bug in our code because a custom enumerator thought that code after the `yield` was guaranteed. This puts a warning to make sure developers are aware of this issue.

### DIFF
--- a/guides/custom-enumerator.md
+++ b/guides/custom-enumerator.md
@@ -72,3 +72,5 @@ end
 ```
 
 We recommend that you read the implementation of the other enumerators that come with the library (`CsvEnumerator`, `ActiveRecordEnumerator`) to gain a better understanding of building Enumerator objects.
+
+Code that is written after the `yield` in a custom enumerator is not guaranteed to execute. In the case that a job is forced to exit ie `job_should_exit?` is true, then the job is re-enqueued during the yield and the rest of the code in the enumerator does not run. You can follow that logic [here](https://github.com/Shopify/job-iteration/blob/9641f455b9126efff2214692c0bef423e0d12c39/lib/job-iteration/iteration.rb#L128-L131).


Mangara (approved):

etiennebarrie:
That's true, but it shouldn't be really surprising: there's no way the gem can give the control back to the code in the enumerator to continue running after the `yield` without also letting it run until the next `yield`. Put in another way, the gem can't communicate to the Enumerator to stop its iteration after a yield.

But since we use `Enumerator#each` and not a bare `Enumerator#next` in a loop, you can simply use an `ensure` section in your Enumerator, and that will always run when `each` exits, even if the enumerator hasn't been exhausted. To reuse the example from the gem StripeListEnumerator:
```ruby  def each
    @list.each do |item, _index|
      yield item, item.id
      puts item.id # this won't run when the job is interrupted
    ensure
      puts item.id # this will always run after each iteration, even if the job is interrupted
    end
  end
```

grcooper:
I had a misunderstanding about ensure, I didn't realize that it also ran after `return` statements as well.

The situation that is happening in the iterator is:

```ruby
def foo
    10.times do |i|
        return if i == 5
    ensure
        puts "bar #{i}"
    end
end

foo
```

Which outputs:

```ruby
bar 0
bar 1
bar 2
bar 3
bar 4
bar 5
```

TIL ensure even runs after a return, but then exits that loop. I had thought because it was not going to be run in general because of the return, then it shouldn't run after the function is returned from.

@etiennebarrie Do you have a suggestion here? Do you think that we should put the `ensure` example  of the `StripeListEnumerator` alongside this warning?

etiennebarrie:
Yes maybe, I can take a look if you want.

PR #61, opened 2020-12-14
Call `super` in `method_added`

paracycle:
The `JobIteration::Iteration` module is not being a good citizen by not calling `super` in the `method_added` hook it is registering. This prevents other modules that might have registered `method_added` hooks from receiving the callback. For example, Sorbet runtime does this, and [`JobIteration::Iteration` breaks its functionality][1].

[1]: https://discourse.shopify.io/t/you-called-sig-twice-without-declaring-a-method-inbetween/12595/2?u=ufuk.kayserilioglu

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -22,6 +22,7 @@ module JobIteration
     module ClassMethods
       def method_added(method_name)
         ban_perform_definition if method_name.to_sym == :perform
+        super
       end
 
       def on_start(*filters, &blk)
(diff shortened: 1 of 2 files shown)


shardulbee:
Is it worth adding a [test?](https://github.com/Shopify/job-iteration/blob/master/test/unit/iteration_test.rb)

paracycle:
> Is it worth adding a [test?](https://github.com/Shopify/job-iteration/blob/master/test/unit/iteration_test.rb?rgh-link-date=2020-12-15T14%3A27%3A57Z)

Oh yeah, right. Good call, I'll add a test.

shardulbee (approved):

shardulbee:
@Shopify/job-patterns could you take a look please?

Mangara (approved):

PR #62, opened 2020-12-18
updated docs to use ApplicationJob instead of ActiveJob::Base

cindygshopify:
updated the README.md to use ApplicationJob instead of ActiveJob::Base as ApplicationJob is not the explicit base class for jobs in Rails 5.


gmcgibbon (approved):

gmcgibbon:
cc @Shopify/job-patterns

grcooper (approved):

BoGs (approved):

GustavoCaso (approved):

PR #64, opened 2021-01-20
Account for drop in enumerable size

sambostock:
When building an `Array` `Enumerator`, we attach the size. Presumably, this size should reflect if we've dropped any elements, which it currently doesn't.

This updates the size calculation to subtract any dropped elements.

I wanted to add an accompanying test (or consult existing tests to see if this is expected behaviour), but I was unable to find anywhere we test `Array` `Enumerator`s, other than incidentally in some of the `ThrottleEnumerator` tests. 🤔

### DIFF
--- a/lib/job-iteration/enumerator_builder.rb
+++ b/lib/job-iteration/enumerator_builder.rb
@@ -59,7 +59,7 @@ module JobIteration
           cursor + 1
         end
 
-      wrap(self, enumerable.each_with_index.drop(drop).to_enum { enumerable.size })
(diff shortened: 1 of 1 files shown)


sambostock:
Test failures against edge `rails/rails` are unrelated. See #65 for details.

sambostock:
Upon further reading, it looks like everywhere else we report the size of the `Enumerator` as the total size of the original collection, not the size "left to enumerate", so the original code is correct. Whoops! 🤷

PR #67, opened 2021-02-19
Migrate travis.yml to github actions workflow

adrianna-chang-shopify:
Figured it might be nice to migrate CI from travis to a GitHub actions workflow. The workflow should be pretty much the same - I've excluded ruby 2.5 + the edge Gemfile, and ruby 2.6 + the edge Gemfile since ruby 2.7 is now a requirement.

Redis needed to be added as a service, but MySQL comes installed on the Ubuntu VM so we can start it directly on the runner.


.github/workflows/ruby.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    name: Ruby ${{ matrix.ruby }}
+    services:
+      redis:
+        image: redis
+        ports:
+        - 6379:6379
+    strategy:
+      matrix:
+        ruby: [2.5, 2.6, 2.7]
```
etiennebarrie, line 16: What about 3.0?

etiennebarrie (approved):

PR #74, opened 2021-04-17
Add Gemfile to GitHub Action Workflow label

sambostock:
Currently there appear to be duplicate workflows running. This is because the workflow only lists the Ruby version in its name.

However, the matrix also uses different `Gemfile`s. By adding those to the name, it becomes evident there are is no duplication.

This also renames the workflow from `Ruby` to `CI` to clean up the label.

Before|After Gemfile Label Addition|After Workflow Rename
-|-|-
image|image|image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Ruby
+name: CI
 
 on: [push]
 
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -5,7 +5,7 @@ on: [push]
 jobs:
   build:
     runs-on: ubuntu-latest
-    name: Ruby ${{ matrix.ruby }}
+    name: Ruby ${{ matrix.ruby }} | Gemfile ${{ matrix.gemfile }}
     services:
       redis:
         image: redis


sambostock:
I did notice that the job name is a little redundant when combined with the workflow name

> Ruby / Ruby $RUBY_VERSION | Gemfile $GEMFILE_NAME

Would it make sense to go with something like

> Ruby / v$RUBY_VERSION | Gemfile $GEMFILE_NAME

or am I just splitting hairs? 😅

tgwizard (approved):
LGTM as is. If you want to shorten/deduplicate it, perhaps it'd be nice to remove the `Ruby / ` prefix?

sambostock:
I believe the names are generated as

```
$WORKFLOW_NAME / $JOB_NAME
```

So we'd have to change the name of the workflow from "Ruby" to something else.

adrianna-chang-shopify (approved):
LGTM! 🌟 

> So we'd have to change the name of the workflow from "Ruby" to something else.

We could rename the workflow to `CI` to make this slightly less redundant? I'm not hugely bothered by it, so I'll defer to you on whether you'd like to change this!

PR #81, opened 2021-04-21
Deprecate usage of cursors not safe to serialize

### LINKED ISSUE #80: Reevaluate rollout of enforced cursor serializability
## Context

The enforcement that cursors be losslessly serializable to JSON introduced in #73 was released in `1.1.11`.

Unfortunately, this broke some production jobs, due to excessive mocking in tests not catching unserializable cursors.

We have since reverted the change in #77 and published version `1.1.12`.

This was discussed in Slack:

- Apps may be relying on the existing behaviour of being able to provide arbitrary cursor and dealing with deserializing the corrupted cursor.
- Those apps would now encounter an error
- This is therefore a breaking change, and would likely require a major version upgrade
- Tests do not always discover breakages, so it may be desirable to be able to customize the behaviour in the event of an unserializable cursor.

## Way Forward

We could reintroduce the cursor validation, but this time with behaviour customizable by the host app.

- By default, we could simply log a warning via `Rails.logger.warn`.
- Host apps wishing to be more strict could select alternative behaviour:
  ```ruby
  JobIteration.configure do |config|
    config.on_cursor_unsafe_to_serialize = :raise
  end
  ```
- Host apps requiring even more customized behaviour could register a callable
  ```ruby
  JobIteration.configure do |config|
    config.on_cursor_unsafe_to_serialize = lambda do |job, cursor|
      # e.g. compare job against some list and conditionally raise
      # e.g. emit a metric to some metric service
    end
  end
  ```

This could be shipped as part of a minor version bump, as it would be a non-breaking new feature intended to provide more robust guarantees.

When we decide it is appropriate to change the default, we can simply change it and release a major version bump. Consumers could continue to override the config until such a time as we decide to get rid of it.

sambostock:
## TL;DR

This is a redo of #73, but using a deprecation warning instead of raising an error. The error is deferred until `2.0`.

## Details

It is unsafe to use cursors which cannot be serialized using JSON.dump and deserialized again using JSON.parse. Such cursors may result in a different cursor being deserialized, or in serialization failing entirely, which may result in incorrect behaviour in iterative jobs.

Starting in 2.0, usage of such cursors will raise an ArgumentError. Until then, a deprecation warning will be logged.

#73 contains more context on the change, and #80 contains information as to the motivation for this alternative approach.

----

Closes #80

### DIFF
--- a/lib/job-iteration.rb
+++ b/lib/job-iteration.rb
@@ -9,6 +9,8 @@ module JobIteration
 
   INTEGRATIONS = [:resque, :sidekiq]
 
+  Deprecation = ActiveSupport::Deprecation.new("2.0", "JobIteration")
+
   extend self
 
   # Use this to _always_ interrupt the job after it's been running for more than N seconds.
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -6,28 +6,6 @@ module JobIteration
   module Iteration
     extend ActiveSupport::Concern
 
-    class CursorError < ArgumentError
-      attr_reader :cursor
-
-      def initialize(message, cursor:)
-        super(message)
-        @cursor = cursor
-      end
-
-      def message
-        "#{super} (#{inspected_cursor})"
-      end
-
-      private
-
-      def inspected_cursor
-        cursor.inspect
-      rescue NoMethodError
-        # For those brave enough to try to use BasicObject as cursor. Nice try.
-        Object.instance_method(:inspect).bind(cursor).call
-      end
-    end
-
     included do |_base|
       attr_accessor(
         :cursor_position,
@@ -142,8 +120,7 @@ module JobIteration
       arguments = arguments.dup.freeze
       found_record = false
       enumerator.each do |object_from_enumerator, index|
-        # Deferred until 2.0.0
-        # assert_valid_cursor!(index)
+        assert_valid_cursor!(index)
 
         record_unit_of_work do
           found_record = true
@@ -206,11 +183,17 @@ module JobIteration
     def assert_valid_cursor!(cursor)
       return if serializable?(cursor)
 
-      raise CursorError.new(
-        "Cursor must be composed of objects capable of built-in (de)serialization: " \
-        "Strings, Integers, Floats, Arrays, Hashes, true, false, or nil.",
-        cursor: cursor,
-      )
+      Deprecation.warn(<<~DEPRECATION_MESSAGE.chomp)
+        Cursor must be composed of objects capable of built-in (de)serialization:
+          Strings, Integers, Floats, Arrays, Hashes, true, false, or nil.
+        #{self.class.name}#build_enumerator's Enumerator provided:
+          #{begin
+              cursor.inspect
+            rescue
+              Object.instance_method(:inspect).bind(cursor).call
+            end}
+        This will raise starting in version 2.0 of JobIteration!
+      DEPRECATION_MESSAGE
     end
 
     def assert_implements_methods!
--- a/test/unit/iteration_test.rb
+++ b/test/unit/iteration_test.rb
@@ -66,6 +66,7 @@ class JobIterationTest < IterationUnitTest
   class InvalidCursorJob < ActiveJob::Base
     include JobIteration::Iteration
     def each_iteration(*)
+      return if Gem::Version.new(JobIteration::VERSION) < Gem::Version.new("2.0")
       raise "Cursor invalid. This should never run!"
     end
   end
@@ -182,58 +183,68 @@ class JobIterationTest < IterationUnitTest
     assert_includes(methods_added, :foo)
   end
 
-  def test_jobs_using_time_cursor_will_raise
-    skip("Deferred until 2.0.0")
+  def test_jobs_using_time_cursor_is_deprecated
     push(JobWithTimeCursor)
-    assert_raises_cursor_error { work_one_job }
+    assert_cursor_deprecation_warning { work_one_job }
   end
 
-  def test_jobs_using_active_record_cursor_will_raise
-    skip("Deferred until 2.0.0")
+  def test_jobs_using_active_record_cursor_is_deprecated
     refute_nil(Product.first)
     push(JobWithActiveRecordCursor)
-    assert_raises_cursor_error { work_one_job }
+    assert_cursor_deprecation_warning { work_one_job }
   end
 
-  def test_jobs_using_symbol_cursor_will_raise
-    skip("Deferred until 2.0.0")
+  def test_jobs_using_symbol_cursor_is_deprecated
     push(JobWithSymbolCursor)
-    assert_raises_cursor_error { work_one_job }
+    assert_cursor_deprecation_warning { work_one_job }
   end
 
-  def test_jobs_using_string_subclass_cursor_will_raise
-    skip("Deferred until 2.0.0")
+  def test_jobs_using_string_subclass_cursor_is_deprecated
     push(JobWithStringSubclassCursor)
-    assert_raises_cursor_error { work_one_job }
+    assert_cursor_deprecation_warning { work_one_job }
   end
 
-  def test_jobs_using_basic_object_cursor_will_raise
-    skip("Deferred until 2.0.0")
+  def test_jobs_using_basic_object_cursor_is_deprecated
     push(JobWithBasicObjectCursor)
-    assert_raises_cursor_error { work_one_job }
+    assert_cursor_deprecation_warning { work_one_job }
   end
 
-  def test_jobs_using_complex_but_serializable_cursor_will_not_raise
-    skip("Deferred until 2.0.0")
+  def test_jobs_using_complex_but_serializable_cursor_is_not_deprecated
     push(JobWithComplexCursor)
-    work_one_job
+    assert_no_cursor_deprecation_warning do
+      work_one_job
+    end
   end
 
   private
 
-  def assert_raises_cursor_error(&block)
-    error = assert_raises(JobIteration::Iteration::CursorError, &block)
-    inspected_cursor = begin
-                         error.cursor.inspect
-                       rescue NoMethodError
-                         Object.instance_method(:inspect).bind(error.cursor).call
-                       end
-    assert_equal(
-      "Cursor must be composed of objects capable of built-in (de)serialization: " \
-      "Strings, Integers, Floats, Arrays, Hashes, true, false, or nil. " \
-      "(#{inspected_cursor})",
-      error.message,
-    )
+  def assert_cursor_deprecation_warning
+    original_behaviour = JobIteration::Deprecation.behavior
+    warning_count = 0
+    prefix = <<~PREFIX
+      DEPRECATION WARNING: Cursor must be composed of objects capable of built-in (de)serialization:
+        Strings, Integers, Floats, Arrays, Hashes, true, false, or nil.
+      #{ActiveJob::Base.queue_adapter.enqueued_jobs.first.fetch("job_class")}#build_enumerator's Enumerator provided:
+    PREFIX
+    JobIteration::Deprecation.behavior = lambda do |message, _callstack, deprecation_horizon, gem_name|
+      warning_count += 1
+      suffix = "This will raise starting in version #{deprecation_horizon} of #{gem_name}!"
+      assert_match(/\A#{Regexp.escape(prefix)}  .+\n#{Regexp.escape(suffix)}/, message)
+    end
+    yield
+    assert_equal(1, warning_count, "expected deprecation warning")
+  ensure
+    JobIteration::Deprecation.behavior = original_behaviour
+  end
+
+  def assert_no_cursor_deprecation_warning
+    original_behaviour = JobIteration::Deprecation.behavior
+    JobIteration::Deprecation.behavior = lambda do |message, _callstack, _deprecation_horizon, _gem_name|
+      flunk("Expected no deprecation warning: #{message}")
+    end
+    yield
+  ensure
+    JobIteration::Deprecation.behavior = original_behaviour
   end
 
   def push(job, *args)
--- a/test/integration/integration_behaviour.rb
+++ b/test/integration/integration_behaviour.rb
@@ -35,7 +35,7 @@ module IntegrationBehaviour
     end
 
     test "unserializable corruption is prevented" do
-      skip "Deferred until 2.0.0"
+      skip "Breaking change deferred until 2.0" if Gem::Version.new(JobIteration::VERSION) < Gem::Version.new("2.0")
       # Cursors are serialized as JSON, but not all objects are serializable.
       #     time   = Time.at(0).utc   # => 1970-01-01 00:00:00 UTC
       #     json   = JSON.dump(time)  # => "\"1970-01-01 00:00:00 UTC\""


lib/job-iteration.rb
```diff
@@ -9,6 +9,8 @@ module JobIteration
 
   INTEGRATIONS = [:resque, :sidekiq]
 
+  Deprecation = ActiveSupport::Deprecation.new("2.0", "JobIteration")
```
sambostock, line 12: I originally used `ActiveSupport::Deprecation.warn` directly, that labelled deprecations as coming from Rails. This uses our own object, but does mean that the host app's settings for `ActiveSupport::Deprecation` do not propagate (e.g. `disallowed_deprecations`).

lib/job-iteration/iteration.rb
```diff
@@ -206,11 +183,17 @@ def build_enumerator(params, cursor:)
...
-      )
+      Deprecation.warn(<<~DEPRECATION_MESSAGE.chomp)
+        Cursor must be composed of objects capable of built-in (de)serialization:
+          Strings, Integers, Floats, Arrays, Hashes, true, false, or nil.
+        #{self.class.name}#build_enumerator's Enumerator provided:
+          #{begin
+              cursor.inspect
+            rescue
+              Object.instance_method(:inspect).bind(cursor).call
+            end}
```
sambostock, line 194: Rubocop didn't like my inline `rescue`. 🤷 

This is so we can handle any cursor, including `BasicObject.new`. When that blows up, it won't be us. 😅

GustavoCaso: If we believe showing the `cursor` as part of the error message is worth it. We could move this into a private method

```ruby
def cursor_to_s
  cursor.inspect
rescue
  Object.instance_method(:inspect).bind(cursor).call
end

Comment (GustavoCaso) [reply lib/job-iteration/iteration.rb:194]: Being honest maybe just mentioning the class of the job that fails is enough for the developer to figure out where to look and what to fix.

If someone is using a custom cursor and gets this error message:

```
Cursor must be composed of objects capable of built-in (de)serialization:
  Strings, Integers, Floats, Arrays, Hashes, true, false, or nil.
#{self.class.name}#build_enumerator's Enumerator provided an invalid cursor object.
```

Could be more than enough

File: lib/job-iteration/iteration.rb [hunk 3]
```diff
@@ -206,11 +183,17 @@ def build_enumerator(params, cursor:)
...
       return if serializable?(cursor)
 
-      raise CursorError.new(
-        "Cursor must be composed of objects capable of built-in (de)serialization: " \
-        "Strings, Integers, Floats, Arrays, Hashes, true, false, or nil.",
-        cursor: cursor,
-      )
+      Deprecation.warn(<<~DEPRECATION_MESSAGE.chomp)
+        Cursor must be composed of objects capable of built-in (de)serialization:
+          Strings, Integers, Floats, Arrays, Hashes, true, false, or nil.
```
GustavoCaso, line 188: Should we also allow to pass Symbol as part of the cursor?

Yesterday we noticed that having a Hash with keys as symbols also broke. Hash with symbols should be accepted as well 😄

sambostock: No. Symbols are not safe to serialize as JSON, because they are deserialized as Strings:

```sh
$ ruby -rjson -e 'p JSON.parse(JSON.dump(:symbol))'
"symbol"
```

`cursor_position` is serialized by the queue adapter (Resque/Sidekiq), not by ActiveJob itself, so none of Rails' magic symbol serialization happens.

test/unit/iteration_test.rb
```diff
@@ -182,58 +183,68 @@ def foo
...
-                         Object.instance_method(:inspect).bind(error.cursor).call
-                       end
-    assert_equal(
-      "Cursor must be composed of objects capable of built-in (de)serialization: " \
-      "Strings, Integers, Floats, Arrays, Hashes, true, false, or nil. " \
-      "(#{inspected_cursor})",
-      error.message,
-    )
+  def assert_cursor_deprecation_warning
+    original_behaviour = JobIteration::Deprecation.behavior
```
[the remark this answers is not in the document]
sambostock: We could, I just wanted to make sure the deprecation message was formatted correctly.

sambostock:
I'll simplify the deprecation warning message 👍

lib/job-iteration/iteration.rb
GustavoCaso: Thanks for the explanation

sambostock:
I've circled back to this and made some changes. This PR still introduces the deprecation warning, however it also introduces a config to force the new behavior. This config can be set globally and/or per-job, allowing consumers to globally opt-in to enforcement ahead of 2.0, and optionally opt-out per job until they are made compliant, or alternatively opt-in per compliant job until they are able to opt-in globally.

This should allow for a smoother transition. Version 2.0 would simply delete (or no-op & deprecate) the config and instruct consumers to remove it from their code.

lib/job-iteration.rb
```diff
@@ -20,6 +22,25 @@ module JobIteration
...
+  # true, false, or nil.
+  #
+  #    JobIteration.enforce_serializable_cursors = true
+  #
+  # For more granular control, this can also be configured per job class, and
+  # is inherited by child classes.
+  #
+  #     class MyJob < ActiveJob::Base
+  #       include JobIteration::Iteration
+  #       self.enforce_serializable_cursors = false
```
sambostock, line 36: @Mangara do you think we should prefix this config with `job_iteration_` like in #240? It seems less likely to collide than `max_job_runtime`, but maybe it makes sense to be consistent?

lib/job-iteration.rb
```diff
@@ -20,6 +22,25 @@ module JobIteration
   # Defaults to nil which means that jobs will not be interrupted except on termination signal.
   attr_accessor :max_job_runtime
 
+  # Set this to `true` to enforce that cursors be composed of objects capable
+  # of built-in (de)serialization: Strings, Integers, Floats, Arrays, Hashes,
+  # true, false, or nil.
```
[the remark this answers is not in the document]
sambostock: Hmm... Good question. If Active Job performs a serialization step before handing off the job to the backend, and if we can test that an object is serializable by it, then that should work. I shall have to look into this.

Indeed my goal is not to be restrictive, but to prevent foot-guns, as my team encountered.

sambostock:
Alright, I've updated the branch again, this time to leverage `ActiveJob::Arguments` to support cursors of any type it supports.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - [241](https://github.com/Shopify/job-iteration/pull/241) - Require Ruby 2.7+, dropping 2.6 support
 - [241](https://github.com/Shopify/job-iteration/pull/241) - Require Rails 6.0+, dropping 5.2 support
+- [80](https://github.com/Shopify/job-iteration/pull/80) - Deprecate un(de)serializable cursors
```
sambostock, line 5: Should probably update this to "Serialize cursors"

lib/job-iteration/iteration.rb
```diff
@@ -288,6 +261,44 @@ def handle_completed(completed)
...
+      # Backwards compatibility for jobs interrupted before cursor serialization feature shipped
+      return job_data["cursor_position"] unless job_data.key?("serialized_cursor_position")
+
+      deserialize_cursor(job_data.fetch("serialized_cursor_position"))
+    end
+
+    def serialize_cursor(cursor)
+      ActiveJob::Arguments.serialize([cursor]).first
+    rescue ActiveJob::SerializationError
+      raise if job_iteration_enforce_serializable_cursors
```
sambostock, line 274: This used to wrap the error in a custom `CursorError`, but I figure it might be adequate to just re-raise the `ActiveJob::SerializationError`.

lib/job-iteration/iteration.rb
```diff
@@ -288,6 +261,44 @@ def handle_completed(completed)
...
+
+      Deprecation.warn(<<~DEPRECATION_MESSAGE)
+        The cursor stored for #{self.class.name} could not be deserialized properly. Falling back to legacy behavior.
+
+        This will raise starting in version #{Deprecation.deprecation_horizon} of #{Deprecation.gem_name}!"
+      DEPRECATION_MESSAGE
+
+      cursor
```
sambostock, line 299: I should add a test for this somehow.

test/support/jobs.rb
```diff
@@ -15,11 +15,12 @@ def each_iteration(omg)
   end
 end
 
-class TimeCursorJob < ActiveJob::Base
+class UnserializableCursorJob < ActiveJob::Base
   include JobIteration::Iteration
+  UnserializableCursor = Class.new
```
sambostock, line 20: Since we're getting all of `ActiveJob`'s serialization now, we have to explicitly create an object that cannot be serialized. `BasicObject` doesn't work though, because that straight up blows up `ActiveJob` because it doesn't respond to `respond_to?` 😅

test/unit/iteration_test.rb
```diff
@@ -199,41 +141,100 @@ def foo
...
   end
 
-  def test_jobs_using_time_cursor_will_raise
-    skip_until_version("2.0.0")
-    push(JobWithTimeCursor)
-    assert_raises_cursor_error { work_one_job }
+  UnserializableCursor = Class.new
+  SerializableCursor = Class.new
+
+  class SerializableCursorSerializer < ActiveJob::Serializers::ObjectSerializer
```
sambostock, line 147: Alternatively, I could make it work with `GlobalID`.

lib/job-iteration/iteration.rb
sambostock: I realized we should never store an unserialized cursor as `serialized_cursor_position`. I'm refactoring so that we only store `serialized_cursor_position` if it succeeded, and only deserialize it if it is there (falling back to unserialized `cursor_position` otherwise).

lib/job-iteration/iteration.rb
```diff
@@ -288,6 +286,25 @@ def handle_completed(completed)
       raise "Unexpected thrown value: #{completed.inspect}"
     end
 
+    def cursor_position_from_job_data(job_data)
+      if job_data.key?("serialized_cursor_position")
+        deserialize_cursor(job_data.fetch("serialized_cursor_position"))
```
sambostock, line 291: I've updated this so if a serialized cursor is found, it should always be deserializable.

test/unit/iteration_test.rb
```diff
@@ -199,41 +156,169 @@ def foo
...
+      work_one_job do |job_data|
+        job_data["cursor_position"] = "should be ignored"
+        job_data["serialized_cursor_position"] = ActiveJob::Arguments.serialize([SerializableCursor.new]).first
+      end
+
+      assert_instance_of(SerializableCursor, InfiniteCursorLoggingJob.cursors.last)
+    end
+  ensure
+    InfiniteCursorLoggingJob.cursors.clear
+  end
```
sambostock, line 308: I'm not sure if I'm missing something, but I wasn't able to find any tests that actually exercise resuming jobs. Everything seems to be testing running or interrupting jobs, but no tests actually resume anything. So this happens to add test coverage for that 😅

guides/custom-enumerator.md
```diff
@@ -82,6 +88,58 @@ class StripeJob < ActiveJob::Base
...
+`FancyCursor`, and Job Iteration would fallback to the legacy behavior of not
+serializing the cursor. This would typically result in the queue adapter
+eventually serializing the cursor as JSON by calling `.to_s` on it. Upon
+deserialization, the cursor would be corrupted as `:abc`, rather than
+`FancyCursor.new(:abc)`.
+
+To avoid this, job authors should take care to ensure that their cursor is
+serializable by Active Job. This can be done in a couple ways, such as:
+- [adding GlobalID support to the cursor class](https://edgeguides.rubyonrails.org/active_job_basics.html#globalid)
+- [implementing a custom serializer for the cursor class](https://edgeguides.rubyonrails.org/active_job_basics.html#serializers)
```
bdewater, line 132:
```suggestion
- [implementing a custom Active Job argument serializer for the cursor class](https://guides.rubyonrails.org/active_job_basics.html#serializers)
```

lib/job-iteration/iteration.rb
```diff
@@ -209,16 +202,21 @@ def build_enumerator(params, cursor:)
...
 
-      raise CursorError.new(
-        "Cursor must be composed of objects capable of built-in (de)serialization: " \
-          "Strings, Integers, Floats, Arrays, Hashes, true, false, or nil.",
-        cursor: cursor,
-      )
+      Deprecation.warn(<<~DEPRECATION_MESSAGE, caller_locations(3))
+        The Enumerator returned by #{self.class.name}#build_enumerator yielded a cursor which is unsafe to serialize.
+
+        TBD MENTION SERIALIZERS!
```
bdewater, line 214:
```suggestion
        See https://github.com/Shopify/job-iteration/blob/main/guides/custom-enumerator.md#cursor-types
```

this will work once this is merged :)

guides/custom-enumerator.md
```diff
@@ -82,6 +88,58 @@ class StripeJob < ActiveJob::Base
...
+- handling (de)serialization in the job/enumerator itself
+    ```diff
```
bdewater, line 134:
```suggestion
    ```ruby
```

bdewater (approved):
Some suggestions to polish things up, but overall this looks good!

sambostock:
Sidekiq 7 has been released, and it now errors if job arguments contain objects that don't round trip as JSON, so this PR is now not only desirable to avoid bugs, but also improves compatibility with Sidekiq.

I'll see if I can get it over the finish line this week.

sambostock:
Looks like CI is broken for Rails edge due to the removal of `ActiveRecord::SchemaMigration.table_name`, as described in
DatabaseCleaner/database_cleaner#693.

guides/custom-enumerator.md
```diff
@@ -82,6 +88,58 @@ class StripeJob < ActiveJob::Base
...
+If this job was interrupted, Active Job would be unable to serialize
+`FancyCursor`, and Job Iteration would fallback to the legacy behavior of not
+serializing the cursor. This would typically result in the queue adapter
+eventually serializing the cursor as JSON by calling `.to_s` on it. Upon
+deserialization, the cursor would be corrupted as `:abc`, rather than
+`FancyCursor.new(:abc)`.
+
+To avoid this, job authors should take care to ensure that their cursor is
+serializable by Active Job. This can be done in a couple ways, such as:
+- [adding GlobalID support to the cursor class](https://edgeguides.rubyonrails.org/active_job_basics.html#globalid)
```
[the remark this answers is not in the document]
sambostock: Nope, that's probably just what I'd gotten to from Google. 🤷 Fixed!

test/unit/iteration_test.rb
```diff
@@ -199,41 +158,169 @@ def foo
...
+        job_data.delete("serialized_cursor_position")
+      end
+
+      assert_equal("raw cursor", InfiniteCursorLoggingJob.cursors.last)
+    end
+  ensure
+    InfiniteCursorLoggingJob.cursors.clear
+  end
+
+  def test_job_interrupted_with_serialized_cursor_position_should_ignore_unserialized_cursor_position
```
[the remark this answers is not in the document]
sambostock: Replied in https://github.com/Shopify/job-iteration/pull/81#discussion_r1398498867

lib/job-iteration/iteration.rb
```diff
@@ -286,6 +281,13 @@ def job_iteration_max_job_runtime
       [global_max, class_max].min
     end
 
+    def job_iteration_enforce_serializable_cursors? # TODO: Add a test for the edge case of registering it afterwards
```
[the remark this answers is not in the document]
sambostock: Good catch – yes, `test_global_max_job_runtime_with_updated_value`

lib/job-iteration/iteration.rb
```diff
@@ -222,16 +214,19 @@ def build_enumerator(params, cursor:)
       EOS
     end
 
-    # The adapter must be able to serialize and deserialize the cursor back into an equivalent object.
-    # https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple
     def assert_valid_cursor!(cursor)
-      return if serializable?(cursor)
+      serialize_cursor(cursor)
+      true
```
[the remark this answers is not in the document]
sambostock: Actually, I think that's wrong, and I was just returning `true`/`false`. Removed.

lib/job-iteration/iteration.rb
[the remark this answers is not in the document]
Mangara: Then we should remove the TODO 🙂

lib/job-iteration/iteration.rb
```diff
@@ -222,16 +214,16 @@ def build_enumerator(params, cursor:)
...
-          "Strings, Integers, Floats, Arrays, Hashes, true, false, or nil.",
-        cursor: cursor,
-      )
+      serialize_cursor(cursor)
+    rescue ActiveJob::SerializationError
+      raise if job_iteration_enforce_serializable_cursors?
+
+      Deprecation.warn(<<~DEPRECATION_MESSAGE, caller_locations(3))
+        The Enumerator returned by #{self.class.name}#build_enumerator yielded a cursor which is unsafe to serialize.
+        See https://github.com/Shopify/job-iteration/blob/main/guides/custom-enumerator.md#cursor-types
```
sambostock, line 224: This needs to match the heading we added.

```suggestion
        See https://github.com/Shopify/job-iteration/blob/main/guides/custom-enumerator.md#cursor-serialization
```

test/unit/iteration_test.rb
```diff
@@ -199,41 +158,169 @@ def foo
...
+  def test_jobs_using_unserializable_cursor_will_not_raise_if_enforce_serializable_cursors_unset_per_class
+    with_global_enforce_serializable_cursors(true) do
+      job_class = build_invalid_cursor_job(cursor: UnserializableCursor.new)
+      job_class.job_iteration_enforce_serializable_cursors = false
+
+      assert_cursor_deprecation_warning_on_perform(job_class)
+    end
+  end
+
+  def test_jobs_using_unserializable_cursor_when_interrupted_should_only_store_the_cursor_and_no_serialized_cursor
```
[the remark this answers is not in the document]
sambostock: Yeah. 2.0 should be able to safely stop reading (and writing) to the old argument, and we can (at least) drop all the tests that use it as an input/output, if not drop all the tests on the implementation and just test the resumption behavior.

PR #82, opened 2021-05-04
Add additional tests for multiple columns sorting

berkcaputcu:
I wanted highlight a hidden gotcha with the use of the `columns` option with the `ActiveRecordEnumerator` and start a discussion around how to handle this gotcha when communicating this to the developers who would want to use this option.

In short, the usage of the `columns` option may end up with these 2 unexpected behaviours:
1. Iteration API may yield the same record twice
2. Iteration API may skip yielding some records that are in the original scope

These unexpected behaviours could happen if the records that are in the scope get updated during the iteration lifecycle. If this update changes the value of the column that the enumerator is sorting by, this update also changes the order of this record. If this value becomes larger than the current cursor position, the iteration API yields this record again at a later time. And the iteration API skips a record, if the value becomes less than the current cursor position.

In my opinion, the usage of `columns` breaks some promises that the iteration API makes and I believe this should be at least documented so that the developers using this feature are not surprised later.

### DIFF
--- a/test/unit/active_job_iteration_test.rb
+++ b/test/unit/active_job_iteration_test.rb
@@ -468,6 +468,54 @@ module JobIteration
       assert_equal(first_products, MultipleColumnsActiveRecordIterationJob.records_performed)
     end
 
+    def test_multiple_columns_iterates_over_records_more_than_once
+      minimum_timestamp = Product.minimum(:updated_at)
+      maximum_timestamp = Product.maximum(:updated_at)
+
+      record = Product.first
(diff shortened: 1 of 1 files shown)


grcooper (approved):

Mangara:
I agree that this should be documented somewhere. That said, tests are not a very useful form of documentation in my opinion. This is not desired behaviour, more of a side-effect of the way this is implemented.

I would prefer a section on the `columns` option in this guide: https://github.com/Shopify/job-iteration/blob/master/guides/iteration-how-it-works.md

PR #84, opened 2021-05-05
adjust_total_time before running on_complete callbacks

### LINKED ISSUE #83: total_time is inaccurate when accessed inside on_complete callback
### Minimal example:

```ruby
    class TestInterruptibleJob < ActiveJob::Base

      def build_enumerator(cursor:)
        enumerator_builder.build_times_enumerator(3, cursor: cursor) # iterate 3 times
      end

      def each_iteration(*)
        sleep(1) # each iteration takes 1 second
      end

     on_complete do
        total_time >= 3 # => false (total_time = 0.0)
     end

    end
```

### Expected:

```ruby
total_time >= 3 # => true
```

### Workaround

```ruby
     on_complete do
        # calculate total_time using [iteration_]start_time and current time.
        total_time_before_on_complete = total_time + (Time.now.utc.to_f - start_time.to_f).round(6)
        total_time_before_on_complete >= 3 # => true
     end
```

pedropb:
Closes #83

### What are you trying to accomplish?

We were using the `on_complete` callback to report the job total run time but we noticed that `total_time` only gets updated after the callbacks are called.

### What approach did you choose and why?

There were 3 options:
1. Move `adjust_total_time` so that it's called before the callbacks.
2. Copy `adjust_total_time` so that it's called before the callbacks and also right before logging the output summary. This could be useful in case the "total_time after callbacks" is valuable.
3. Publish a new event via `ActiveSupport::Notifications.instrument` instead. This would overlap with the callback functionality.

I decided for 1 because I think it's the one that makes the most sense.

### What reviewers should focus on?

- I'm not familiar with writing RubyGems so I added Timecop as a dependency without a group. If that's problematic I can try to come up with another solution to stub the passage of time.
- I would consider this a bugfix but if users are relying on the inaccurate `total_time` inside `on_complete` callbacks, this patch would be a breaking change.

### DIFF
--- a/test/unit/iteration_test.rb
+++ b/test/unit/iteration_test.rb
@@ -121,6 +121,23 @@ class JobIterationTest < IterationUnitTest
     end
   end
 
+  class JobThatCompletesAfter3Seconds < ActiveJob::Base
+    include JobIteration::Iteration
+    include ActiveSupport::Testing::TimeHelpers
+    def build_enumerator(assertions, cursor:)
+      @assertions = assertions
+      enumerator_builder.build_times_enumerator(3, cursor: cursor) # iterate 3 times
+    end
+
+    def each_iteration(*)
+      travel(1.second) # each iteration takes 1 second
+    end
+
+    on_complete do
+      @assertions.call(self)
+    end
+  end
+
   def test_jobs_that_define_build_enumerator_and_each_iteration_will_not_raise
     push(JobWithRightMethods, "walrus" => "best")
     work_one_job
@@ -217,6 +234,12 @@ class JobIterationTest < IterationUnitTest
     skip("Deferred until 2.0.0")
     push(JobWithComplexCursor)
     work_one_job
+  end
+
+  def test_jobs_using_on_complete_have_accurate_total_time
+    freeze_time do
+      JobThatCompletesAfter3Seconds.perform_now(->(job) { assert_equal(3, job.total_time) })
+    end
   end
 
   private
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -162,6 +162,8 @@ module JobIteration
         "times_interrupted=#{times_interrupted} cursor_position=#{cursor_position}"
       ) unless found_record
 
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -27,3 +27,5 @@ gem "rake"
 
 # for unit testing optional sorbet support
 gem "sorbet-runtime"
+
+gem "timecop", "~> 0.9.4"
```
Mangara, line 31: Could we use the [ActiveSupport time helpers](https://api.rubyonrails.org/v5.2.5/classes/ActiveSupport/Testing/TimeHelpers.html) instead? The gem already depends on parts of Rails, so that feels like a smaller step than a completely separate gem.

pedropb:
Latest pushed changes:
- Moved `adjust_total_time` from the `completed?` branch to the end of `iterate_with_enumerator`.
- Dropped TimeCop in favour of `ActiveSupport::Testing::TimeHelpers`.
- Consolidate the test files by merging the separate `callbacks_test.rb` logic into `iteration_test.rb`.

test/unit/iteration_test.rb
```diff
@@ -219,6 +236,12 @@ def test_jobs_using_complex_but_serializable_cursor_will_not_raise
     work_one_job
   end
 
+  def test_jobs_using_on_complete_have_accurate_total_time
+    freeze_time do
+      JobThatCompletesAfter3Seconds.perform_now(->(job) { assert_equal(3, job.total_time) })
+    end
+  end
```
GustavoCaso, line 243: I would suggest simplifying the test:

```ruby
def test_jobs_using_on_complete_have_accurate_total_time
    freeze_time do
      job = JobThatCompletesAfter3Seconds.new
      job.perform_now
      assert_equal(3, job.total_time)
    end
  end
```

That way you do not need to pass the assertion around, which is a little weird 😄

GustavoCaso (commented):
Code LGTM. I would like it if you could modify the test file to remove the need to pass the assertions around. Apart from that this is ready to be merge 😄

GustavoCaso (approved):

PR #85, opened 2021-05-07
Add a section about risks of using columns argument

berkcaputcu:
When the columns argument is used with the
`build_active_record_enumerator`, the enumerator may
skip or yield records multiple times.

I've highlighted how these can happen in this PR: https://github.com/Shopify/job-iteration/pull/82

@grcooper @Mangara I decided to add this comment to the only place where it talks about the `columns` option, instead of here: https://github.com/Shopify/job-iteration/blob/master/guides/iteration-how-it-works.md 

Also, it's a bit challenging to explain this edge case in a few sentences, so please help me clarify it better if it looks unclear.

Let me know what you think about it.


Mangara (approved):
This change is fine, but I don't think many people will see it. We should document usage of the `cursor` option outside the code, and include these warnings. That said, I fully understand if you don't want to be the one tackling that 🙂 

This change is going in the right direction, so I'm 👍 with merging it.

PR #86, opened 2021-05-07
Adds `as_relation` flag to `EnumeratorBuilder#active_record_on_batches`

adrianna-chang-shopify:
Currently, the resulting batch from `#active_record_on_batches` is an Array of records. Many job authors end up converting this into a relation to perform batch operations such as `#update_all` (even the existing `BatchesJob` example in the README does this!). By expanding the API of `#active_record_on_batches` to return a relation if a flag is set, users who require their batch as an `ActiveRecord::Relation` will no longer need to perform an additional query to turn records back into a relation.

This has no effect on existing jobs, since by default `as_relation` will be set to false and batches will continue to be converted to Arrays.

Not sure whether it's worth documenting an example of this in the README, or whether the API documentation is sufficient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,28 @@ class BatchesJob < ApplicationJob
 
   def each_iteration(batch_of_comments, product_id)
     # batch_of_comments will contain batches of 100 records
-    Comment.where(id: batch_of_comments.map(&:id)).update_all(deleted: true)
+    batch_of_comments.each do |comment|
+      DeleteCommentJob.perform_later(comment)
+    end
+  end
+end
+```
+
+```ruby
+class BatchesAsRelationJob < ApplicationJob
+  include JobIteration::Iteration
+
+  def build_enumerator(product_id, cursor:)
+    enumerator_builder.active_record_on_batches_as_relation(
+      Product.find(product_id).comments,
+      cursor: cursor,
+      batch_size: 100,
+    )
+  end
+
+  def each_iteration(batch_of_comments, product_id)
+    # batch_of_comments will be a Comment::ActiveRecord_Relation
+    batch_of_comments.update_all(deleted: true)
   end
 end
 ```
--- a/lib/job-iteration/active_record_cursor.rb
+++ b/lib/job-iteration/active_record_cursor.rb
@@ -48,14 +48,7 @@ module JobIteration
       @position = position
     end
 
-    def update_from_record(record)
-      self.position = @columns.map do |column|
-        method = column.to_s.split(".").last
-        record.send(method.to_sym)
-      end
-    end
-
-    def next_batch(batch_size)
+    def next_batch(batch_size, as_relation = false)
       return nil if @reached_end
 
       relation = @base_relation.limit(batch_size)
@@ -64,14 +57,11 @@ module JobIteration
         relation = relation.where(*conditions)
       end
 
-      records = relation.uncached do
-        relation.to_a
+      if as_relation
+        prepare_next_batch_as_relation(relation)
+      else
+        prepare_next_batch_as_records(batch_size, relation)
       end
-
-      update_from_record(records.last) unless records.empty?
-      @reached_end = records.size < batch_size
-
-      records.empty? ? nil : records
     end
 
     protected
@@ -92,6 +82,37 @@ module JobIteration
       ret = @position.reduce([conditions]) { |params, value| params << value << value }
       ret.pop
       ret
+    end
+
+    private
+
+    def prepare_next_batch_as_records(batch_size, relation)
+      records = relation.uncached do
+        relation.to_a
+      end
+
+      update_from_record(records.last) unless records.empty?
+      @reached_end = records.size < batch_size
+
+      records.empty? ? nil : records
+    end
+
+    def prepare_next_batch_as_relation(relation)
+      last_record = relation.last
+      if last_record
+        update_from_record(last_record)
+        relation
+      else
+        @reached_end = true
+        nil
+      end
+    end
+
+    def update_from_record(record)
+      self.position = @columns.map do |column|
+        method = column.to_s.split(".").last
+        record.send(method.to_sym)
+      end
     end
   end
 end
--- a/lib/job-iteration/active_record_enumerator.rb
+++ b/lib/job-iteration/active_record_enumerator.rb
@@ -6,9 +6,10 @@ module JobIteration
   class ActiveRecordEnumerator
     SQL_DATETIME_WITH_NSEC = "%Y-%m-%d %H:%M:%S.%N"
 
-    def initialize(relation, columns: nil, batch_size: 100, cursor: nil)
+    def initialize(relation, columns: nil, batch_size: 100, as_relation: false, cursor: nil)
       @relation = relation
       @batch_size = batch_size
+      @as_relation = as_relation
       @columns = Array(columns || "#{relation.table_name}.#{relation.primary_key}")
       @cursor = cursor
     end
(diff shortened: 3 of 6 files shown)


etiennebarrie (approved):

adrianna-chang-shopify:
>  Would migrate batches as a relation by default, would mean that it would be a breaking change, no?

Oh yes absolutely, this would definitely require a major version update & changes to the jobs in Core (although most of these jobs want a relation, so we'd be reducing db calls here at least).

You're right, I think having the two options is fine for now. If we decide we want to make relations the default moving forward, it will indeed be easier to have users migrate to a new API first rather than flipping a default on an existing API. 

I'll make the changes, and see if I can reduce some of those queries with the optimization suggestions Étienne made.

adrianna-chang-shopify:
+1 to @etiennebarrie's point, although the flip side of that is that if we're bothering to make batches relations by default to remove that extra call, we'd obviously want to go in there and remove any `.where(id: batch.map(&:id))` calls to actually reap the benefits of switching to a relation. So there would still be changes involved.

If our end goal is to come back to a single API for active record batch enumerators, I think a flag would make more sense because we could:
1) Ship this with the flag; 2) Change the jobs in Core to use `as_relation: true` and remove excess db calls 3) Ship a major version that defaults the flag to true

But to @GustavoCaso's point, I'd rather ship an additional API and unblock the work in the Maintenance Tasks gem, and defer any decisions about yielding batches as relations by default until a further date.

adrianna-chang-shopify:
👋  @GustavoCaso sorry for the delay on this, we made a couple of changes! Notably:
- Optimized the logic in `#next_batch` when returning relations to avoid calling `#size`, which was producing an extra `COUNT` db query as @etiennebarrie mentioned [here](https://github.com/Shopify/job-iteration/pull/86#discussion_r630245401)
      - Note that before, each batch was performing a SELECT (`records.last`) and a COUNT (`records.empty`), and even if we removed the `records.empty`, we had a `records.size` that would end up doing a COUNT
      - Now, we call `next_batch` _one extra time_  (once with an empty batch to set `@reached_end`, and then again where we return early)
      - `next_batch` on records still works the same, and since we can use `#size` on the records without a db call, we can set `@reached_end` based on this, without needing to do an extra `next_batch` call like with records
      - Tests for all this logic are here: https://github.com/Shopify/job-iteration/pull/86/commits/1d0c5708dee9d299a352772e751fe86a034dd874#diff-51044c0cc0b5d8318cedf297cb91496754e8fb32f2820b44cf031ee04ce5c5f0
      - I also made `#update_from_record` private, I don't see a reason why it needs public visibility
 - Separate enumerator builder for relations as you'd suggested, `#active_record_on_batches_as_relation`
 - I've opted to keep it as a flag rather than splitting out Cursor classes for now - I think it makes sense to eventually separate out different ActiveRecord cursors, but I don't think it necessarily makes sense for the relation one to be a subclass of the records one (I'd prefer for them both to subclass an abstract Active Record Cursor class). I think there's also room to split `ActiveRecordEnumerator` into multiple enumerators as well - one for records and one for batches? Anyways, TL;DR is I think there's room for some refactoring in general, and I don't want to commit to defining an abstraction yet. So, I think it makes sense to leave as a flag for now

Let me know what you think!

adrianna-chang-shopify:
@etiennebarrie I could be missing something, but seems to me that
https://github.com/Shopify/job-iteration/blob/bae9b71987bba9e4dc03e3e60c3e1ab28013f0ff/lib/job-iteration/active_record_enumerator.rb#L31

_doesn't_ actually load everything again.

We're doing `relation.last` here in `#prepare_next_batch_as_relation`: https://github.com/Shopify/job-iteration/pull/86/files#diff-a62e3d57d0bb79088cbdb20f8f1480e8422c2b02222057de2624ed0c1d4eac82R101, which ends up calling `load` on the relation and setting `@records` 

So even though we're doing:
https://github.com/rails/rails/blob/75ac626c4e21129d8296d4206a1960563cc3d4aa/activerecord/lib/active_record/relation/finder_methods.rb#L147-L148 and then using [`#find_last`](https://github.com/rails/rails/blob/75ac626c4e21129d8296d4206a1960563cc3d4aa/activerecord/lib/active_record/relation/finder_methods.rb#L574-L576) because we have a limit on our relation, this ends up using `records.last`, and in [`#records`](https://github.com/rails/rails/blob/v6.1.3.2/activerecord/lib/active_record/relation.rb#L248-L251) when we [`#load`](https://github.com/rails/rails/blob/v6.1.3.2/activerecord/lib/active_record/relation.rb#L636-L642) it won't actually perform any more db queries because we already are `loaded?`.

Similarly, `any?` should be fine because we're also loaded, so it'll just use `@records.empty?` without another query.

etiennebarrie:
Yes you got that right but what that means, is that even using relations, the records are loaded, which is basically the same as calling `to_a`.
There's a difference, because doing `relation.update_all` will re-use the previous conditions on the relation, including the limit, whereas doing `Record.where(id: records.map(&:id)).update_all` will use the ids as the condition to update records, but the end result is still that we loaded the full fields of 100 records from the DB and created active record objects with them.

For example we could test this by changing our test to:

```ruby
    def test_activerecord_batches_as_relation
      push(BatchActiveRecordRelationIterationJob)

      work_one_job
      assert_jobs_in_queue(0)

      records_performed = BatchActiveRecordIterationJob.records_performed
      assert_equal([3, 3, 3, 1], records_performed.map(&:size))
      assert(records_performed.all? { |relation|
        relation.is_a?(ActiveRecord::Relation) && !relation.loaded?
      })
    end
```

You can also see this by logging the SQL queries:
```diff
diff --git c/test/unit/active_record_enumerator_test.rb i/test/unit/active_record_enumerator_test.rb
index f5a17ee..494f5a7 100644
--- c/test/unit/active_record_enumerator_test.rb
+++ i/test/unit/active_record_enumerator_test.rb
@@ -140,7 +140,7 @@ def build_enumerator(relation: Product.all, batch_size: 2, columns: nil, cursor:
     def count_uncached_queries(&block)
       count = 0
 
-      query_cb = ->(*, payload) { count += 1 unless payload[:cached] }
+      query_cb = ->(*, payload) { p payload.slice(:sql, :cached); count += 1 unless payload[:cached] }
       ActiveSupport::Notifications.subscribed(query_cb, "sql.active_record", &block)
       count
     end
```

You'll see that even using the relations, we load the records:
```
{:sql=>"SELECT `products`.* FROM `products` ORDER BY products.id LIMIT 2"}
{:sql=>"SELECT `products`.* FROM `products` WHERE (products.id > 2) ORDER BY products.id LIMIT 2"}
{:sql=>"SELECT `products`.* FROM `products` WHERE (products.id > 4) ORDER BY products.id LIMIT 2"}
{:sql=>"SELECT `products`.* FROM `products` WHERE (products.id > 6) ORDER BY products.id LIMIT 2"}
{:sql=>"SELECT `products`.* FROM `products` WHERE (products.id > 8) ORDER BY products.id LIMIT 2"}
{:sql=>"SELECT `products`.* FROM `products` WHERE (products.id > 10) ORDER BY products.id LIMIT 2"}
```

When we would expect completely different queries, e.g. something like
```sql
SELECT `products`.* FROM `products` ORDER BY products.id OFFSET 1 LIMIT 1
SELECT `products`.* FROM `products` WHERE (products.id > 2) ORDER BY products.id OFFSET 1 LIMIT 1
SELECT `products`.* FROM `products` WHERE (products.id > 4) ORDER BY products.id OFFSET 3 LIMIT 1
```

i.e. loading the last record of the batch only.

adrianna-chang-shopify:
We are going to take a different approach of building a new enumerator so that we can pluck only the cursor columns and don't have to load full records in order to compute the cursor value. This will also allow us to return a relation based on PK ids, rather than the original relation, which in some cases may be quite complex. We'll open this on a separate branch 😁

PR #87, opened 2021-05-13
Optimize edge case in throttle enumerator

vahe:
Currently when using the throttle enumerator, even when throttled, the first value of the passed enumerator is evaluated.
This may be problematic in cases when we're throttling based on database health, as we will still perform an ActiveRecord query when throttled.

This PR updates the throttle enumerator to check the throttler before evaluating the next value.

### DIFF
--- a/lib/job-iteration/throttle_enumerator.rb
+++ b/lib/job-iteration/throttle_enumerator.rb
@@ -34,7 +34,7 @@ module JobIteration
             throw(:abort, :skip_complete_callbacks)
           end
 
-          yielder.yield(*@enum.next)
+          yielder.yield(*@enum.next_values)
         end
       end
     end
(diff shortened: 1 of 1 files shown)


hyhuang1218 (approved):
👍

test/unit/throttle_enumerator_test.rb
```diff
@@ -110,6 +110,24 @@ def each_iteration(record, _params)
       assert_equal 0, IterationThrottleJob.on_complete_called
     end
 
+    test "does not evaluate enumerator when throttled" do
+      class FlunkOnIterationThrottleJob < ActiveJob::Base
```
sambostock, line 114: Looks good except we should avoid be declaring constants (such as classes) in test bodies.

Can we move this to the test class body instead?

sambostock (changes requested):

test/unit/throttle_enumerator_test.rb
vahe: Fixed

PR #91, opened 2021-05-19
Batch Enumerator Yielding Relations

adrianna-chang-shopify:
Take two of #86 

## Context
- Currently, the resulting batch from `#active_record_on_batches` is an array of records
- Many job authors ultimately want a relation (ie. to perform a batch operation such as `#update_all` or `#delete_all`) and convert this to a relation: `Model.where(id: records.map(&:id)).update_all`. This produces an extra query
- Not only do we perform this extra query, but we've already loaded the full records into memory, when instead we could optimize and only pluck the cursor-related columns to produce the cursor

## Proposed Solution
- New API for returning an enumerator yielding batch relations: `#active_record_on_batch_relations`, leaving existing batch enumerator API for records intact
- We're written this as a new enumerator class, `ActiveRecordBatchEnumerator`. The existing `ActiveRecordEnumerator` enumerator and `ActiveRecordCursor` classes have a lot of duplication and should probably be refactored. I intend to go back and abstract away cursor-related details from this new `ActiveRecordBatchEnumerator` class, while also fixing up the existing `ActiveRecordCursor`, in a separate PR.
- This new class is an actual `Enumerator` and defines `#each`
- This new batch enumerator has two key optimizations to note:
     - Instead of using `record = relation.last` and then querying the cursor columns on the `record` to construct the cursor, [we pluck _only_ the cursor columns](https://github.com/Shopify/job-iteration/compare/batch-enumerator?expand=1#diff-910f147f7f9b7040713e971f5b348b6085fc95f263730453546d7c4a6b4bf553R73). `relation.last` will load all of the records because we are using a `LIMIT`, as described [here](https://github.com/Shopify/job-iteration/pull/86#issuecomment-839894209). Consequently, we optimize by only plucking what we need.
     - We build the relation to yield using the [primary keys of the records in the original relation](https://github.com/Shopify/job-iteration/compare/batch-enumerator?expand=1#diff-910f147f7f9b7040713e971f5b348b6085fc95f263730453546d7c4a6b4bf553R64). Since we have these values anyways, building a relation based on the PKs ends up being much more efficient for the user when they load the relation inside `#each_iteration` than reusing the original relation (which may have had complex logic / joins / etc). This was actually inspired by how Rails does [`#in_batches`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation/batches.rb#L230-L238)
-  The rest of the logic is pretty similar to what's in `ActiveRecordEnumerator` and `ActiveRecordCursor`, with some simplifications, given that now everything is happening within a single object.

## End Result
- Users can enumerate on relations directly
- These relations are not loaded when they are yielded
- We perform a single query per batch, `SELECT <cursor_columns> FROM relation LIMIT <batch_size>`

### DIFF
--- a/lib/job-iteration/active_record_batch_enumerator.rb
+++ b/lib/job-iteration/active_record_batch_enumerator.rb
@@ -60,8 +60,7 @@ module JobIteration
       @cursor = Array.wrap(cursor)
 
       # Yields relations by selecting the primary keys of records in the batch.
-      # Post.where(published: nil)
-      # results in an enumerator of relations like Post.where(ids: batch_of_ids)
+      # Post.where(published: nil) results in an enumerator of relations like: Post.where(ids: batch_of_ids)
       @base_relation.where(@primary_key => ids)
     end
 
@@ -72,9 +71,10 @@ module JobIteration
       end
 
       column_values = relation.pluck(*@pluck_columns)
-      primary_key_values = column_values.map { |values| values[@primary_key_index || -1] }
+      primary_key_index = @primary_key_index || -1
+      primary_key_values = column_values.map { |values| values[primary_key_index] }
 
-      column_values.map! { |values| values.map! { |value| column_value(value) } }
+      serialize_column_values!(column_values)
       [column_values, primary_key_values]
     end
 
@@ -99,6 +99,10 @@ module JobIteration
       ret = @cursor.reduce([where_clause]) { |params, value| params << value << value }
       ret.pop
       ret
+    end
+
+    def serialize_column_values!(column_values)
+      column_values.map! { |values| values.map! { |value| column_value(value) } }
     end
 
     def column_value(value)
--- a/lib/job-iteration/active_record_batch_enumerator.rb
+++ b/lib/job-iteration/active_record_batch_enumerator.rb
@@ -1,0 +1,108 @@
+# frozen_string_literal: true
+
+module JobIteration
+  # Builds Batch Enumerator based on ActiveRecord Relation.
+  # @see EnumeratorBuilder
+  class ActiveRecordBatchEnumerator
+    include Enumerable
+
+    SQL_DATETIME_WITH_NSEC = "%Y-%m-%d %H:%M:%S.%N"
+
+    def initialize(relation, columns: nil, batch_size: 100, cursor: nil)
+      @batch_size = batch_size
+      @primary_key = "#{relation.table_name}.#{relation.primary_key}"
+      @columns = Array(columns&.map(&:to_s) || @primary_key)
+      @primary_key_index = @columns.index(@primary_key) || @columns.index(relation.primary_key)
+      @pluck_columns = if @primary_key_index
+        @columns
+      else
+        @columns.dup << @primary_key
+      end
+      @cursor = Array.wrap(cursor)
+      raise ArgumentError, "Must specify at least one column" if @columns.empty?
+      if relation.joins_values.present? && !@columns.all? { |column| column.to_s.include?(".") }
+        raise ArgumentError, "You need to specify fully-qualified columns if you join a table"
+      end
+
+      if relation.arel.orders.present? || relation.arel.taken.present?
+        raise ConditionNotSupportedError
+      end
+
+      @base_relation = relation.reorder(@columns.join(","))
+    end
+
+    def each
+      while (relation = next_batch)
+        break if @cursor.empty?
+        yield relation, cursor_value
+      end
+    end
+
+    def size
+      @base_relation.count
+    end
+
+    private
+
+    def next_batch
+      relation = @base_relation.limit(@batch_size)
+      if conditions.any?
+        relation = relation.where(*conditions)
+      end
+
+      cursor_values, ids = relation.uncached do
+        pluck_columns(relation)
+      end
+
+      cursor = cursor_values.last
+      # The primary key was plucked, but original cursor did not include it
+      cursor.pop unless @primary_key_index || cursor.nil?
+      @cursor = Array.wrap(cursor)
+
+      # Yields relations by selecting the primary keys of records in the batch.
+      # Post.where(published: nil)
+      # results in an enumerator of relations like Post.where(ids: batch_of_ids)
+      @base_relation.where(@primary_key => ids)
+    end
+
+    def pluck_columns(relation)
+      if @pluck_columns.size == 1 # only the primary key
+        column_values = relation.pluck(*@pluck_columns)
+        return [column_values, column_values]
+      end
+
+      column_values = relation.pluck(*@pluck_columns)
+      primary_key_values = column_values.map { |values| values[@primary_key_index || -1] }
+
+      column_values.map! { |values| values.map! { |value| column_value(value) } }
+      [column_values, primary_key_values]
+    end
+
+    def cursor_value
+      return @cursor.first if @cursor.size == 1
+      @cursor
+    end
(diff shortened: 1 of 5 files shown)


adrianna-chang-shopify:
cc @Shopify/rails @Shopify/job-patterns

casperisfine (approved):

adrianna-chang-shopify:
@etiennebarrie shall we ship this?

test/unit/active_record_batch_enumerator_test.rb
```diff
@@ -0,0 +1,103 @@
...
+      expected_product_cursor = [products.last.updated_at.strftime(SQL_TIME_FORMAT), products.last.id]
+      assert_equal([products, expected_product_cursor], enum.first)
+    end
+
+    test "columns configured with primary key only queries primary key column once" do
+      queries = track_uncached_queries do
+        enum = build_enumerator(columns: [:updated_at, :id])
+        enum.first
+      end
+      assert_equal 1, queries.first.scan(/`products`.`id`/).count
```
etiennebarrie, line 60: This is testing the search for the primary key existing as part of the columns here, right?
https://github.com/Shopify/job-iteration/blob/0090f5de24a26f8470bffe43746c09a11ce522b1/lib/job-iteration/active_record_batch_enumerator.rb#L15
Maybe we could check what the selected columns are instead?
```suggestion
      assert_equal "`products`.`updated_at`, `products`.`id`", queries.first[/SELECT (.*) FROM/, 1]
```

test/unit/active_record_batch_enumerator_test.rb
```diff
@@ -0,0 +1,103 @@
...
+      cursor = [products.last.created_at.strftime(SQL_TIME_FORMAT), products.last.id]
+      assert_equal([products, cursor], enum.first)
+
+      enum = build_enumerator(columns: [:created_at, :id], cursor: cursor)
+      products = Product.order(:created_at, :id).offset(2).take(2)
+
+      cursor = [products.last.created_at.strftime(SQL_TIME_FORMAT), products.last.id]
+      assert_equal([products, cursor], enum.first)
+    end
+
```
etiennebarrie, line 84: Shouldn't we test `size`?

test/unit/active_job_iteration_test.rb
```diff
@@ -448,6 +462,18 @@ def test_activerecord_batches
...
     end
 
+    def test_activerecord_batch_relations
+      push(BatchActiveRecordRelationIterationJob)
+
+      work_one_job
+      assert_jobs_in_queue(0)
+
+      records_performed = BatchActiveRecordIterationJob.records_performed
+      assert_equal([3, 3, 3, 1], records_performed.map(&:size))
```
etiennebarrie, line 472: Could we use `count` to make it a bit clearer we're doing a `COUNT(*)` here?

test/unit/active_record_batch_enumerator_test.rb
```diff
@@ -0,0 +1,103 @@
...
+        relation,
+        batch_size: batch_size,
+        columns: columns,
+        cursor: cursor,
+      )
+    end
+
+    def track_uncached_queries(&block)
+      queries = []
+      query_cb = ->(*, payload) { queries << payload[:sql] unless payload[:cached] }
```
etiennebarrie, line 98: I think we don't need the cached/uncached distinction anymore, since we don't do cached queries anymore?

test/unit/active_record_batch_enumerator_test.rb
```diff
@@ -0,0 +1,103 @@
...
+    def build_enumerator(relation: Product.all, batch_size: 2, columns: nil, cursor: nil)
+      JobIteration::ActiveRecordBatchEnumerator.new(
+        relation,
+        batch_size: batch_size,
+        columns: columns,
+        cursor: cursor,
+      )
+    end
+
+    def track_uncached_queries(&block)
```
etiennebarrie, line 96: Maybe we could add another test using this, just pointing out that we do one query per yield + one at the end to get the empty cursor?
We do show in the job iteration test that we yield relations with 3 + 3 + 3 + 1 records, but not that we do 5 queries in that case (1 per yield + a final one to get the empty cursor).

lib/job-iteration/active_record_batch_enumerator.rb
```diff
@@ -0,0 +1,112 @@
...
+        relation = relation.where(*conditions)
+      end
+
+      cursor_values, ids = relation.uncached do
+        pluck_columns(relation)
+      end
+
+      cursor = cursor_values.last
+      # The primary key was plucked, but original cursor did not include it
+      cursor.pop unless @primary_key_index || cursor.nil?
```
etiennebarrie, line 59: The `cursor.nil?` part here is when we reached the end and `cursor_values` is empty, right?
Maybe we could skip the `Array.wrap` as well, and nil-check instead of empty-check in `each`?

etiennebarrie (approved):
Some propositions.

lib/job-iteration/active_record_batch_enumerator.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+
+      if relation.arel.orders.present? || relation.arel.taken.present?
+        raise ConditionNotSupportedError
+      end
+
+      @base_relation = relation.reorder(@columns.join(","))
+    end
+
+    def each
```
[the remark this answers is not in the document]
adrianna-chang-shopify: Yeah, I'm also of the mindset that `Wrapper` is not very useful. Plus, `build_throttle_enumerator` just returns a Ruby Enumerator directly without wrapping it, and this was implemented more recently (~1 year ago), so I think it makes sense for us to do the same.

lib/job-iteration/active_record_batch_enumerator.rb
```diff
@@ -32,9 +32,13 @@ def initialize(relation, columns: nil, batch_size: 100, cursor: nil)
...
-      while (relation = next_batch)
-        break if @cursor.empty?
-        yield relation, cursor_value
+      if block_given?
+        while (relation = next_batch)
+          break if @cursor.empty?
+          yield relation, cursor_value
+        end
+      else
+        to_enum(:each)
```
etiennebarrie, line 41: You forgot the size here 😄

lib/job-iteration/active_record_batch_enumerator.rb
```diff
@@ -0,0 +1,117 @@
...
+        raise ConditionNotSupportedError
+      end
+
+      @base_relation = relation.reorder(@columns.join(","))
+    end
+
+    def each
+      if block_given?
+        while (relation = next_batch)
+          break if @cursor.nil?
```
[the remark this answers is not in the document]
rafaelfranca: That is not clear at all. You are relying in hoisting to define the value of a variable and you need to remember that is how the interpreter works. In that case it would be better to do:

```ruby
      cursor = cursor_values.last
      if cursor.present?
        # The primary key was plucked, but original cursor did not include it, so we should remove it
        cursor.pop unless @primary_key_index
        @cursor = Array.wrap(cursor)
     else
        @cursor = nil
     end
```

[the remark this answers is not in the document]
adrianna-chang-shopify: Revisited the code and realized that `@cursor` actually can't be `nil` 🤦‍♀️  The value will carry over from the previous batch, even if we return early. The check is unnecessary, so we'll put something up to remove it.

Issue #89, opened 2021-05-17
Support Faktory


plasticine:
Hey there 👋 we’re very keen to investigate adopting `job-iteration`, however currently we’re using [Faktory](https://github.com/contribsys/faktory) as our background job implementation.

Is this a patch that the maintainers would be open to? — Happy to take on implementing this if so.

Issue #88, opened 2021-05-17
Support Que


plasticine:
Hey there 👋 we’re very keen to investigate adopting `job-iteration`, however currently we’re using [Que](https://github.com/que-rb/que) as our background job implementation.

Is this a patch that the maintainers would be open to? — Happy to take on implementing this if so.

PR #96, opened 2021-05-27
Infer interruption handler from job

### LINKED ISSUE #90: Support multiple present background worker implementations
For various reasons we support several background job implementations in production — namely [Que](https://github.com/que-rb/que) and [Faktory](https://github.com/contribsys/faktory).

We’re very interested in `job-iteration`, however an immediate issue we face (other than support being added #88 & #89) was the fact that the gem attempts to discover and configure the interruption adapter up-front for a given worker — [here](https://github.com/Shopify/job-iteration/blob/master/lib/job-iteration.rb#L64). Unfortunately for us given our multiple queue implementations in play that share a codebase/monorepo, it means that we’d hit [this branch](https://github.com/Shopify/job-iteration/blob/master/lib/job-iteration.rb#L43-L47), and be stuck. 

I think supporting multiple background job implementations, while perhaps a little unusual, is a very useful use-case to support. Supporting multiple implementations allows for a lot more flexibility around migrating from system to system, or in cases where multiple background systems are desirable in order to place jobs on background systems with differing characteristics.

I would propose that instead of configuring the interruption adapter once during init for a given worker we just discover it around the execution of each iteration, I think this would be fairly achievable and am happy to take a look if this is a patch that would be of interest. 😃

plasticine:
This was a bit of a quick attempt to fix #90

Firstly I wanted to apologise for the size of this patch, and how disruptive the renaming here is—understand this might be a non-starter for various reasons, but I was keen to use autoloading and the current naming scheme breaks this AFAIK.

Happy to try and rework and open to suggestions on how to proceed here if this patch is otherwise of interest. 🙃

If this patch gets accepted I’ll be very keen to follow up with further patches for #88 & #89.


plasticine:
Hey all 👋 — rebased this again, but keen to hear feedback on whether this patch is desirable or not. Very keen to try and get `job-iteration` into a state where we can use it, so happy to rework this into something that will be merged if needs be.

plasticine:
Hey @sambostock @adrianna-chang-shopify 👋 Sorry for a ping on this directly, but I’d be really interested in some feedback here if either of you has a moment (or can redirect this in a better direction).

Mangara:
> It's a little hard to parse the diff because of the rename -- maybe that could get split into a separate commit up front, with the changes to the integration code committed separately?

I'd prefer a separate PR even.

plasticine:
Hey all, thanks very much for the feedback, and apologies again for the messiness of this patch — it got a bit out of hand as I was hacking around. 🙃 

How does the core idea of inferring the integration strike y’all?—If that’s palatable then I’ll rework this patch into something a little less unwieldy without the renaming.

Mangara:
> How does the core idea of inferring the integration strike y’all?

It makes sense to me.

adrianna-chang-shopify:
Makes a lot of sense to me too - since Rails supports specifying the queue adapter on a per-job basis, it seems unnecessarily inflexible to force Job Iteration users to stick to a single queuing backend for their whole application.

PR #97, opened 2021-05-28
Batch enumerator size should return the number of batches, not records

etiennebarrie:
Enumerator#size returns the number of yields, not the underlying number of elements, e.g.:

```ruby
[1,2,3].each_slice(2).size # => 2
```

Similarly BatchEnumerator#size should return the number of batches, not the number of records.

### DIFF
--- a/lib/job-iteration/active_record_batch_enumerator.rb
+++ b/lib/job-iteration/active_record_batch_enumerator.rb
@@ -40,7 +40,7 @@ module JobIteration
     end
 
     def size
-      @base_relation.count
+      (@base_relation.count + @batch_size - 1) / @batch_size # ceiling division
     end
 
     private
--- a/test/unit/active_record_batch_enumerator_test.rb
+++ b/test/unit/active_record_batch_enumerator_test.rb
@@ -35,9 +35,11 @@ module JobIteration
       assert_equal([], enum.to_a)
     end
 
-    test "#size returns size of the base Active Record Relation" do
+    test "#size returns size of the Enumerator" do
       enum = build_enumerator
-      assert_equal Product.count, enum.size
+      assert_equal 5, enum.size # 5 batches of 2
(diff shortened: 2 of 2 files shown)


lib/job-iteration/active_record_batch_enumerator.rb
```diff
@@ -40,7 +40,7 @@ def each
     end
 
     def size
-      @base_relation.count
+      -@base_relation.count.div(-@batch_size)
```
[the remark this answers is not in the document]
etiennebarrie: I found a better way to do the ceiling division without using negative numbers and the fact that divmod/div round the quotient to negative infinity. I find it less confusing.

etiennebarrie:
> Can we update `ActiveRecordEnumerator` to return an enum with the proper size for `#batches` as well?

I don't really want to do it since it would change an existing behavior, potentially breaking things.

adrianna-chang-shopify:
> I don't really want to do it since it would change an existing behavior, potentially breaking things.

Fair, I'd like to circle back to this and make the fix in Core, since technically it's an incorrect implementation of size based on what the enum is yielding.

lib/job-iteration/active_record_batch_enumerator.rb
```diff
@@ -40,7 +40,7 @@ def each
     end
 
     def size
-      @base_relation.count
+      (@base_relation.count + @batch_size - 1) / @batch_size # ceiling division
```
adrianna-chang-shopify, line 43: Actually, leaning back towards using ` (@base_relation.count.to_f / @batch_size).ceil`, then we could probably drop the comment

etiennebarrie: I prefer keeping the computation in integers rather than going to floats. 🤷

PR #105, opened 2021-09-09
Improve dev commands

Mangara:
This PR adds more functionality to `dev test` and adds a `dev style` command to run linting.

New options include `dev test test/unit/iteration_test.rb` to run just the tests in that file, and `dev test test/unit/iteration_test.rb test_that_it_has_a_version_number` to run just the single test.

Non-Shopify contributors can use this same functionality by replacing `dev test` with `bin/test`. For example, `bin/test test/unit/iteration_test.rb` runs just the tests in that file.

All of it uses the existing Rake task under the hood, so contributors familiar with that can keep using it.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -1,0 +1,32 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+def main
+  begin
+    command = create_command
+  rescue ArgumentError => e
+    abort(e.message)
+  end
+  puts "Running #{command.join(" ")}"
+  system(*command)
+end
+
+def create_command
+  case ARGV.length
+  when 0
+    ["bundle", "exec", "rake", "test"]
+  when 1
+    filename = ARGV[0]
+    ["bundle", "exec", "rake", "test", "TEST=#{filename}"]
+  when 2
+    filename = ARGV[0]
+    test_name = ARGV[1]
+    test_name_with_underscores = test_name.tr(" ", "_")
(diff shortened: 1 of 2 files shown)


bin/test
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+def main
+  begin
+    command = create_command
+  rescue ArgumentError => e
+    puts e.message
+    exit
```
sambostock, line 9: Should we exit unsuccessfully? We could do something like
```suggestion
    abort(e.message)
```

bin/test
```diff
@@ -0,0 +1,34 @@
...
+  when 1
+    filename = ARGV[0]
+    "bundle exec rake test TEST=#{filename}"
+  when 2
+    filename = ARGV[0]
+    test_name = ARGV[1]
+    test_name_with_underscores = test_name.tr(" ", "_")
+    test_name_pattern = "/#{Regexp.escape(test_name_with_underscores)}/"
+    test_name_pattern_escaped_single_quotes = test_name_pattern.gsub("'", %q(\\\'))
+    "bundle exec rake test TEST=#{filename} TESTOPTS=\"--name=#{test_name_pattern_escaped_single_quotes} -v\""
```
sambostock, line 28: Would we have to worry about this escaping if we constructed the command as an array to pass to `system`?

sambostock (approved):
Couple small things, but nothing worth blocking.

Issue #102, opened 2021-07-15
Housekeeping Items


sambostock:
There are a couple things we could to improve the gem:

- [ ] Specify a `required_ruby_version` in `gemspec`
  _This makes it clear to consumers which versions of Ruby we commit to supporting._
  - We only test against Ruby 2.5 and greater, so it seems reasonable that it should be at least `>= 2.5`
    - However, the Ruby 2.5 EOL date was 2021-03-31, so maybe we should use `>= 2.6` and stop testing against `2.5`?
- [ ] Bump the Rubocop `TargetRubyVersion` to match the above (it is currently `2.4.4`, whose EOL date was 2020-03-31
- [ ] Remove `Gemfile.lock` from `.gitignore`, as per [this post on the Bundler blog](https://bundler.io/blog/2018/01/17/making-gem-development-a-little-better.html)
  - [ ] Setup Dependabot to keep dependencies up to date.
    - Does Dependabot support our multiple gemfiles?
- [ ] Add internal Shopify owners to the gem

PR #109, opened 2021-09-21
Commit Gemfile.lock to Github

GustavoCaso:
Related to https://github.com/Shopify/job-iteration/issues/102

Commit `Gemfile.lock` following best practices https://bundler.io/blog/2018/01/17/making-gem-development-a-little-better.html

@Shopify/job-platform


Mangara (approved):
We should make sure CI still runs against the latest version and not the version locked here.

PR #147, opened 2021-11-12
Fix enumerator #size for explicit column select

pjambet:
Use `count(:all)` instead of [`count`][2] to get the number of items in
a `ActiveRecordEnumerator` and in a `ActiveRecordBatchEnumerator`. This
prevents SQL syntax errors when the relation has an explicit column
selection with `select(:col1, :col2)`, which would attempt to run the
invalid:

```sql
SELECT COUNT(col1, col2) FROM table;
```

Note that this is very similar to what [ActiveRecord::Relation#size][1]
does, without the "smart" logic to return the length of the array of
records already loaded. This keeps the logic in this gem simpler, and
closer to what we had before with regular `count()`.

The issue surfaced [in maintenance_tasks][3], where `#size` is called
for a `ActiveRecordBatchEnumerator` with tasks not defining a custom
`count` method. This would fail for batches selecting a subset of
columns, such as `Product.select(:id, :name).in_batches`

[1]:https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-size
[2]:https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-count
[3]:https://github.com/Shopify/maintenance_tasks/blob/55a5fd3765dd312ba85eaad47b6e8e81feb566a5/app/jobs/concerns/maintenance_tasks/task_job_concern.rb#L107

### DIFF
--- a/test/unit/active_record_batch_enumerator_test.rb
+++ b/test/unit/active_record_batch_enumerator_test.rb
@@ -48,6 +48,13 @@ module JobIteration
       assert_equal 4, enum.size # 3 batches of 3, 1 batch of 1
     end
 
+    test "#size returns size of the Enumerator with a subset of columns" do
+      enum = build_enumerator(relation: Product.select(:id, :name))
+      assert_equal 5, enum.size # 5 batches of 2
+      enum = build_enumerator(relation: Product.select(:id, :name), batch_size: 3)
+      assert_equal 4, enum.size # 3 batches of 3, 1 batch of 1
+    end
+
     test "batch size is configurable" do
       enum = build_enumerator(batch_size: 4)
       products = Product.order(:id).take(4)
--- a/lib/job-iteration/active_record_batch_enumerator.rb
+++ b/lib/job-iteration/active_record_batch_enumerator.rb
@@ -40,7 +40,7 @@ module JobIteration
     end
 
     def size
-      (@base_relation.count + @batch_size - 1) / @batch_size # ceiling division
+      (@base_relation.count(:all) + @batch_size - 1) / @batch_size # ceiling division
     end
 
     private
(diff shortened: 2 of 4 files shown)


etiennebarrie (commented):
Right, select + count isn't working great indeed.

Though I don't think we should aim for `ActiveRecord::Relation#size` directly. It's smart in that it uses the size of the records array if it's loaded, which it shouldn't be (since it's the base relation that spans all the records you want to iterate over) but still, I'd rather not deal with the smartness here and deal with SQL a bit more directly.
https://github.com/rails/rails/blob/v6.1.4.1/activerecord/lib/active_record/relation.rb#L260
So I think maybe we should use `count(:all)` directly?

test/unit/active_record_batch_enumerator_test.rb
```diff
@@ -48,6 +48,13 @@ class ActiveRecordBatchEnumeratorTest < IterationUnitTest
       assert_equal 4, enum.size # 3 batches of 3, 1 batch of 1
     end
 
+    test "#size returns size of the Enumerator with a subset of columns" do
+      enum = build_enumerator(relation: Product.select(:id, :name))
+      assert_equal 5, enum.size # 5 batches of 2
+      enum = build_enumerator(relation: Product.select(:id, :name), batch_size: 3)
```
etiennebarrie, line 54: Also `select` for batches is kind of weird because you shouldn't operate on the records, just on the batches (`update_all`/`delete_all`/`destroy_all`) so you shouldn't need to list columns. But we should still fix it 👍

pjambet: Yeah, that's a good point. The issue actually surfaced when using batches, and using an `update_all` call, but using `select` for only the columns we needed as a way to (micro?) optimize the job to really only fetch what we needed from the DB. Probably only realistically saved not that much memory in that one case, but for larger batches, it might.

adrianna-chang-shopify (approved):
Awesome commit message, thank you! ❤️

Mangara (approved):

PR #153, opened 2021-11-17
Reenqueue the job after shutdown hooks have run

MatthewRBruce:
We discovered a race condition where a job can be re-enqueued a new run executed before the previous runs' shutdown hooks have been completed.  This change ensures the shutdown hooks have run before the job is re-enqueued.


lib/job-iteration/iteration.rb
```diff
@@ -132,6 +134,10 @@ def interruptible_perform(*arguments)
 
       run_callbacks(:shutdown)
 
+      if needs_reenqueue && !completed.nil?
```
cardy31, line 137: You can use `completed.present?` here instead of `!completed.nil?`

lib/job-iteration/iteration.rb
```diff
@@ -68,6 +69,7 @@ def ban_perform_definition
 
     def initialize(*arguments)
       super
+      self.needs_reenqueue = false
```
cardy31, line 72: Lets drop the `self` here

lib/job-iteration/iteration.rb
```diff
@@ -141,6 +147,8 @@ def interruptible_perform(*arguments)
     def iterate_with_enumerator(enumerator, arguments)
       arguments = arguments.dup.freeze
       found_record = false
+      self.needs_reenqueue = false
```
cardy31, line 150: `@needs_reenqueue`

lib/job-iteration/iteration.rb
cardy31: Matt and I chatted about this and it breaks the logic since `completed.present?` isn't _quit_ the same as `!completed.nil?`

lib/job-iteration/iteration.rb
```diff
@@ -132,6 +133,10 @@ def interruptible_perform(*arguments)
 
       run_callbacks(:shutdown)
 
+      if @needs_reenqueue && !completed.nil?
```
MatthewRBruce, line 136: Just thinking about this, I think the `completed.nil?` check is pointless.  There can't be a `throw :abort` between when we set `@needs_reenqueue` and the return that's on the next line

cardy31: Agreed. `completed` should be `false` if `@ needs_reenqueue` is true.

cardy31 (approved):
Nice job!

Mangara (approved):

PR #159, opened 2021-11-29
Wrap `active_record_on_batch_relations`

etiennebarrie:
In #91 we forgot to wrap the `ActiveRecordBatchEnumerator`, so setting `wrapper` on the `EnumeratorBuilder` results in a batch enumerator not being wrapped, which is an issue if throttling is implemented with a wrapper for instance.

I realized we had no tests, and no way of making sure potential new enumerator builder methods would be wrapped, so I added a test file for this. It's pretty heavy in metaprogramming but it would have caught our mistake. 🤷 

While doing so I noticed we wrap the enumerators multiple times, which is probably not ideal. `build_times_enumerator` is wrapped twice while `build_once_enumerator` is wrapped three times. We should probably fix that as well at some point.

---

Wrapper is documented in [`JobIteration::EnumeratorBuilder::Wrapper`](https://www.rubydoc.info/gems/job-iteration/1.3.3/JobIteration/EnumeratorBuilder/Wrapper).

Ultimately, I think this ought to be removed from the gem. Users would be better off using replacing the `enumerator_builder` by their own object that delegates or something similar, there's no need for `EnumeratorBuilder` itself to handle that (since they would have to overwrite `enumerator_builder` in their job anyway to change the `wrapper` of the `EnumeratorBuilder` instance).
Based on the comment, there was an idea to check in Iteration that the returned value was wrapped (today we check it's an `Enumerator`) but it never panned out, not sure if it's worth it.


Mangara (approved):
It took me a while to figure out how these tests worked. Perhaps a comment or two would help?

nvasilevski (approved):

PR #160, opened 2021-11-30
Ignore integrations for YARD

etiennebarrie:
It's started failing with the latest YARD because of/thanks to https://github.com/lsegal/yard/issues/1392

cc @Shopify/rails

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -1,3 +1,4 @@
 --no-private
 --embed-mixins
 --markup markdown
+--exclude lib/job-iteration/integrations/.*


.yardopts
```diff
@@ -1,3 +1,4 @@
 --no-private
 --embed-mixins
 --markup markdown
+--exclude lib/job-iteration/integrations/.*
```
etiennebarrie, line 4: These are regexps not globs:
https://github.com/lsegal/yard/blob/v0.9.27/lib/yard/cli/yardoc.rb#L576
https://github.com/lsegal/yard/blob/v0.9.27/lib/yard/parser/source_parser.rb#L105

PR #184, opened 2022-02-10
Release 1.3.5

basicBrogrammer:


basicBrogrammer:
> Can you update the changelog?

Almost forgot 👍  thanks

sambostock:
According to Semantic Versioning, this should have been a minor version update, not a patch, as it introduced a new feature in a backwards compatible way, as opposed to just fixing a bug.

PR #194, opened 2022-03-04
Add Ruby 3.1 and Rails 6.1 and 7 to the CI matrix

petergoldstein:
This PR adds explicit gemfiles for Rails 6.1 and Rails 7.  It updates the Ruby 3.1 configuration to run against these two new entries as well as against Rails edge.

### DIFF
--- a/gemfiles/rails_6_1.gemfile
+++ b/gemfiles/rails_6_1.gemfile
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+eval_gemfile "../Gemfile"
+
+if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new("3.1")
+  gem "net-imap", require: false
+  gem "net-pop", require: false
+  gem "net-smtp", require: false
+end
+
+gem "activejob", "~> 6.1.0"
+gem "activerecord", "~> 6.1.0"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,16 +13,20 @@ jobs:
         - 6379:6379
     strategy:
       matrix:
-        ruby: ["2.6", "2.7", "3.0"]
-        gemfile: [rails_5_2, rails_6_0, rails_edge]
+        ruby: ["2.6", "2.7", "3.0", "3.1"]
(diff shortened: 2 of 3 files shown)


gemfiles/rails_6_1.gemfile
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+eval_gemfile "../Gemfile"
+
+if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new("3.1")
```
Mangara, line 5: Can you explain why this is necessary?

gemfiles/rails_6_1.gemfile
```diff
@@ -0,0 +1,12 @@
...
+  gem "net-imap", require: false
+  gem "net-pop", require: false
+  gem "net-smtp", require: false
```
Mangara, line 8: Should we specify versions for these gems?

Mangara (commented):
Thanks for the PR! We should definitely do this.
I just have some questions about the conditional gems for the 6.1 gemfile.

gemfiles/rails_6_1.gemfile
petergoldstein: The `mail` gem has an issue discussing the need here - https://github.com/mikel/mail/pull/1439 .  And the Rails team discusses it here - https://github.com/rails/rails/pull/44600 .

These gems are not included by default in Ruby 3.1, so they need to be explicitly included.  Attempting to explicitly include them in older Rubies causes issues, so the invoking Gemfile is supposed to explicitly include them.

gemfiles/rails_6_1.gemfile
petergoldstein: I don't think it's required.  I haven't seen any of the fixes for this issue in various repos add versions.

Mangara:
I want to see a CI run for this PR before merging. I think we had an issue in our workflow setup that prevented it from triggering a CI run. That should be corrected after merging https://github.com/Shopify/job-iteration/pull/195, so I'm going to lose and re-open this PR to try and trigger a CI run.

Mangara (approved):

PR #195, opened 2022-03-04
Also run checks when PRs are opened

Mangara:
We received [an external PR](https://github.com/Shopify/job-iteration/pull/194), but CI did not run on it. I suspect it's because the workflow only triggers on `push`, and this PR was from a branch in a fork. This PR adds the `pull_request` trigger, so hopefully CI will run on these type of PRs in the future.


adrianna-chang-shopify (approved):
Once we ship the PR to fix CI ,🚢  !

PR #221, opened 2022-05-13
doc: add example with arguments to `each_iteration

### LINKED ISSUE #220: document parameters to build_enumerator and each_iteration
Currently, the README doesnt explicitely state how to pass additional arguments into the jobs.
I will craft a PR, adding a simple example.

fwolfst:
fixes #220

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,24 @@ class CsvJob < ApplicationJob
 end
 ```
 
+```ruby
+class AdditionalArgumentsJob < ApplicationJob
+  include JobIteration::Iteration
+
+  def build_enumerator(message, name, age, cursor:)
+    enumerator_builder.active_record_on_records(
+      User.where(name: name, age: age),
+      cursor: cursor,
+    )
+  end
+
+  # addtional arguments as in #build_enumerator (except the cursor)
+  def each_iteration(user, message, *)
+    user.notify_about_something(message)
+  end
+end
+```
+
 Iteration hooks into Sidekiq and Resque out of the box to support graceful interruption. No extra configuration is required.
 
 ## Guides


README.md
```diff
@@ -132,6 +132,24 @@ class CsvJob < ApplicationJob
...
+  include JobIteration::Iteration
+
+  def build_enumerator(message, name, age, cursor:)
+    enumerator_builder.active_record_on_records(
+      User.where(name: name, age: age),
+      cursor: cursor,
+    )
+  end
+
+  # addtional arguments as in #build_enumerator (except the cursor)
```
etiennebarrie, line 146: typo here

etiennebarrie (commented):
The `BatchesJob` and `BatchesAsRelationJob` already pass additional arguments. Your added class doesn't really add much to that. Maybe if we just had an additional sentence before the example that would be more explicit. By the way the CSV example should be fixed because it passes an argument to `build_enumerator` which is not passed to `each_iteration`.

fwolfst:
> The `BatchesJob` and `BatchesAsRelationJob` already pass additional arguments. Your added class doesn't really add much to that. Maybe if we just had an additional sentence before the example that would be more explicit. By the way the CSV example should be fixed because it passes an argument to `build_enumerator` which is not passed to `each_iteration`.

Whatever solves #220 . I really find the example and the verbal explanations lacking (for this aspect). Also, my example makes explicit that you might need different stuff to create the enumerator and how to deal with that situation. Should I include the csv and typo-fix and sign the CLA or will this issue not be solved via this PR?

etiennebarrie:
Yes sorry I wasn't clear, this is great, adding more explicit verbal explanation right before your added example is a great idea and I will merge this 👍

PR #226, opened 2022-06-02
Avoid UnknownAttributeReference error by eliminating Array#join

marksiemers:


marksiemers:
Sure thing, what is working using the code in this PR, but was not without this change is the following:
```ruby
  def collection
    return OrdinanceLookup.none unless status.in?(DataSetJobStatus::VALID_STATUSES)

    batch = OrdinanceLookupBatch.find_by!(external_identifier: batch_external_identifier)
    batch.ordinance_lookups
  end
```

What still isn't working is:
```ruby
  def collection
    return OrdinanceLookup.none unless status.in?(DataSetJobStatus::VALID_STATUSES)

    batch = OrdinanceLookupBatch.find_by!(external_identifier: batch_external_identifier)
    batch.ordinance_lookups.in_batches
  end
```

Perhaps it is that the collection is an association?

Do you think it would work to have:
```ruby
batch = OrdinanceLookupBatch.find_by!(external_identifier: batch_external_identifier)
OrdinanceLookup.where(batch: batch)
```

bray:
We're still seeing this. Any update on a fix?

Mangara:
It's not entirely clear to me what this fixes. The code in https://github.com/Shopify/job-iteration/pull/226#issuecomment-1145913181 seems to be for a maintenance task, not a direct job-iteration job. Could you add a test the fails without this change and passes with it?

marksiemers:
I have signed the CLA

PR #240, opened 2022-06-28
Enable per job class `max_job_runtime`

sambostock:
This allows incremental adoption of the setting, without applying the setting globally. 

```ruby
JobIteration.max_job_runtime = nil

class ParentJob < ApplicationJob
  self.max_job_runtime = 5.minutes
  # ...
end
```

Alternatively, it allows applications to set a conservative global setting, and a more aggressive setting per jobs.

```ruby
class ChildJob < ParentJob
  self.max_job_runtime # 5 minutes, inherited
end
class CarefulChildJob < ParentJob
  self.max_job_runtime = 3.minutes # override
end
```

In order to prevent rogue jobs from causing trouble, the per-job override can only be set to a value less than the inherited value.

```ruby
class RogueJob < ParentJob
  self.max_job_runtime = 10.minutes # 💥 ArgumentError
end
class UncleJob < ApplicationJob
  self.max_job_runtime = 10.minutes # ✅ No global default or inherited value, so okay
end
```

### ⚠️ Before Merging
- [x] Clean up commits

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -39,6 +39,33 @@ module JobIteration
       define_callbacks :start
       define_callbacks :shutdown
       define_callbacks :complete
+
+      class_attribute(
+        :job_iteration_max_job_runtime,
+        instance_writer: false,
+        instance_predicate: false,
+        default: JobIteration.max_job_runtime,
+      )
+
+      singleton_class.prepend(PrependedClassMethods)
+    end
+
+    module PrependedClassMethods
+      def job_iteration_max_job_runtime=(new)
+        existing = job_iteration_max_job_runtime
+
+        if existing && (!new || new > existing)
+          existing_label = existing.inspect
+          new_label = new ? new.inspect : "#{new.inspect} (no limit)"
+          raise(
+            ArgumentError,
+            "job_iteration_max_job_runtime may only decrease; " \
+              "#{self} tried to increase it from #{existing_label} to #{new_label}",
+          )
+        end
+
+        super
+      end
     end
 
     module ClassMethods
@@ -262,7 +289,7 @@ module JobIteration
     end
 
     def job_should_exit?
-      if ::JobIteration.max_job_runtime && start_time && (Time.now.utc - start_time) > ::JobIteration.max_job_runtime
+      if job_iteration_max_job_runtime && start_time && (Time.now.utc - start_time) > job_iteration_max_job_runtime
         return true
       end
 
--- a/test/unit/iteration_test.rb
+++ b/test/unit/iteration_test.rb
@@ -251,6 +251,107 @@ class JobIterationTest < IterationUnitTest
     end
   end
 
+  def test_per_class_max_job_runtime_with_default_global
+    freeze_time
+    parent = build_slow_job_class(iterations: 3, iteration_duration: 30.seconds)
+    child = Class.new(parent) do
+      self.job_iteration_max_job_runtime = 1.minute
+    end
+
+    parent.perform_now
+    assert_empty(ActiveJob::Base.queue_adapter.enqueued_jobs)
+
+    child.perform_now
+    assert_partially_completed_job(cursor_position: 2)
+  end
+
+  def test_per_class_max_job_runtime_with_global_set_to_nil
+    freeze_time
+    with_global_max_job_runtime(nil) do
+      parent = build_slow_job_class(iterations: 3, iteration_duration: 30.seconds)
+      child = Class.new(parent) do
+        self.job_iteration_max_job_runtime = 1.minute
+      end
+
+      parent.perform_now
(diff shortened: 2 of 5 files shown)


Mangara (commented):
I really like the idea! Should we namespace the setting on the job, to avoid conflicts? Something like `self.job_iteration_max_runtime`?

sambostock:
> Should we namespace the setting on the job, to avoid conflicts?

This had crossed my mind, and I'm open to it. On the one hand, it makes sense to avoid conflicts. On the other hand, we don't namespace anything else (e.g. `on_start`, `on_shutdown`, & `on_complete` callback DSL methods). I guess it would be possible for Active Job, or some adapter to introduce a similar idea of a time limit... 🤔

lib/job-iteration/iteration.rb
```diff
@@ -39,6 +39,32 @@ def inspected_cursor
...
+      )
+
+      singleton_class.prepend(PrependedClassMethods)
+    end
+
+    module PrependedClassMethods
+      def max_job_runtime=(new)
+        existing = max_job_runtime
+
+        if existing && (!new || new > existing)
```
adrianna-chang-shopify, line 57: Is it worth adding one more test case for the scenario where a global runtime is set, and a class attempts to set `max_job_runtime` to nil, in order to exercise the `if existing && !new` piece of the condition?

test/unit/iteration_test.rb
```diff
@@ -251,6 +251,75 @@ def test_global_max_job_runtime
     end
   end
 
+  def test_per_class_max_job_runtime_with_no_global_set
+    freeze_time
+    with_global_max_job_runtime(nil) do
```
adrianna-chang-shopify, line 256: I find this a bit confusing, because I'd expect "with no global set" to mean that we don't touch `JobIteration.max_job_runtime` at all, not that we explicitly set it to `nil`. If we were to change what not setting the default means at a future point in time (e.g. if `JobIteration.max_job_runtime` defaults to something other than `nil` at a future point), this would fail -- although arguably it would be easy enough to reason about why 😅 

Is there a reason we need to use `with_global_max_job_runtime` here -- can we just rely on what the configured default is, without modifying `JobIteration.max_job_runtime`?

adrianna-chang-shopify (approved):
I really like this as a feature! ❤️ A couple small comments, and I'll leave it to the Job Platform experts to make the call on naming, but this looks good from my perspective!

lib/job-iteration/iteration.rb
```diff
@@ -262,7 +289,7 @@ def output_interrupt_summary
     end
 
     def job_should_exit?
-      if ::JobIteration.max_job_runtime && start_time && (Time.now.utc - start_time) > ::JobIteration.max_job_runtime
+      if job_iteration_max_job_runtime && start_time && (Time.now.utc - start_time) > job_iteration_max_job_runtime
```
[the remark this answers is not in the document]
sambostock: We do [instrument that a job has been interrupted](https://github.com/Shopify/job-iteration/blob/5fddd412af12de37788b9ad7610098df211731d2/lib/job-iteration/iteration.rb#L183), although we do not note the reason why.

I think that could be something worth considering, but I would consider it outside the scope of this PR. I've opened #247 to capture the discussion.

sambostock:
> why this feature is being added now

I've wanted to push for adopting `max_job_runtime` in a number of apps owned by my team for some time now, but we already have a number of existing jobs. Being able to adopt the setting gradually would make that less risky, so I figured I'd implement it.

For example, some of our long jobs run on schedules that mean that they rarely end up interrupted by deploys (middle of the night). We had one such job that was exceptionally interrupted once, only to turn out to improperly deserialize its cursor and blow up every time it tried to resume (the motivation for #73 & #81).

lib/job-iteration/iteration.rb
simbasdad, line 292: Sorry to comment on such an old PR, please let me know if you'd rather I convert this to an issue, or move the conversation somewhere else.

This change actually creates a subtle behaviour change.  Previously, we read the global configuration at job runtime.  Now, we read the global configuration when `JobIteration::Iteration` is loaded, subsequent changes are ignored.

Today, we noticed that maintenance tasks were never pausing because that gem were setting a default job runtime in an `after_initialze` block that was run after this module was included.

I'm not actually sure which option is better.  Just wanted to point out the difference in behaviours.  In the meantime, I've proposed Shopify/maintenance_tasks#918 to fix maintenance tasks under the assumption that the current behaviour of job-iteration is correct.

sambostock: I can try to have a look on Monday, but I think the fix would be to override the method defined by `class_attribute` to delegate to the top level default, as opposed to setting it to the default value at the time the method was defined.

sambostock: #436 should fix this.

PR #241, opened 2022-06-28
Drop support for Ruby 2.6 & Rails 5.2 due to EOL

sambostock:
This drops support for [Ruby 2.6](https://endoflife.date/ruby) and [Rails 5.2](https://endoflife.date/rails), as they are past their end-of-life dates.

CI is failing the most recent version of `rubocop` supports Ruby 2.6 and has breaking configuration changes, and while `rubocop-shopify`'s most recent version addresses the breaking changes, it dropped Ruby 2.6 support several versions ago. Rather than work around this, we can simply drop support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### Master (unreleased)
 
 - [241](https://github.com/Shopify/job-iteration/pull/241) - Require Ruby 2.7+, dropping 2.6 support
+- [241](https://github.com/Shopify/job-iteration/pull/241) - Require Rails 6.0+, dropping 5.2 support
 
 ## v1.3.6 (Mar 9, 2022)
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,12 +14,8 @@ jobs:
     strategy:
       matrix:
         ruby: ["2.7", "3.0", "3.1"]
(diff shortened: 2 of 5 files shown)


dev.yml
```diff
@@ -7,7 +7,7 @@ up:
     - mysql-client:
         or:        [mysql@5.7]
   - ruby:
-      version: 2.6.5
+      version: 3.1.2
```
Mangara, line 10: Should we leave this at 2.7? It'll make it easier to see that you're using a >2.7 feature without having to wait for CI.

Mangara (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ### Master (unreleased)
 
+- [TBD](https://github.com/Shopify/job-iteration/pull/TBD) - Require Ruby 2.7+, dropping 2.6 support
+- [TBD](https://github.com/Shopify/job-iteration/pull/TBD) - Require Rails 6.0+, dropping 5.2 support
```
peterzhu2118, line 4:
```suggestion
- [241](https://github.com/Shopify/job-iteration/pull/241) - Require Ruby 2.7+, dropping 2.6 support
- [241](https://github.com/Shopify/job-iteration/pull/241) - Require Rails 6.0+, dropping 5.2 support
```

PR #250, opened 2022-07-11
Update StripeListEnumerator example

bdewater:
There has been some misunderstanding on how to do rate limiting with cases like these. The job infrastructure isn't the right place for this, the [Stripe gem `auto_paging_each` method](https://github.com/stripe/stripe-ruby/blob/0aca5d2cdce1c734521ab16afe600278a8f1c6f0/lib/stripe/list_object.rb#L51-L86) can make a lot of requests rather quickly before some interruption takes place and this has caused some trouble.

The reason to advise unlimited retries for `MyRateLimiter::LimitExceededError` is because it is expected to raise often and not indicative of a problem. Developers should retry and backoff as usual when errors from the Stripe gem (such as `Stripe::RateLimitError`) are raised - this could mean the internal rate limiter is not configured properly or upstream is having other issues.

### DIFF
--- a/guides/custom-enumerator.md
+++ b/guides/custom-enumerator.md
@@ -31,8 +31,8 @@ class StripeListEnumerator
     pagination_params = {}
     pagination_params[:starting_after] = cursor unless cursor.nil?
 
+    # The following line makes a request, consider adding your rate limiter here.
     @list = resource.public_send(:list, params.merge(pagination_params), options)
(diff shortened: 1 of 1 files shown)


bdewater:
Thanks! Just wanted to make sure this didn't seem completely off base from a job infra perspective :)

I'll be testing it in production soon, if no new learnings come from it I'll merge this.

PR #252, opened 2022-07-12
Investigate using `ActiveJob::Serializers` to serialize cursor

sambostock:
This investigates an alternate approach to #81, as suggested by @bdewater.

If this approach makes sense, #81 will be updated and this PR will be closed.
If not, then it will also be closed.


bdewater (commented):
Thanks for investigating this further. I like the outcome, it plays well with existing Active Job semantics and the only awkward thing (`Arguments.serialize([cursor]).first`) is not that bad & tucked away.

sambostock:
Cool. I'll integrate this work into the other branch. 👍

sambostock:
Closing in favour of #81, as these changes have been integrated into it.

Issue #63, opened 2021-01-04
Support for nested iteration


Mangara:
It is sometimes convenient to iterate over one set of records, and for each of those, over an associated set of records.

As a specific example, in `Shopify/shopify` iterating over all products that satisfy some condition, such as having draft status, is much more efficient if we do it shop by shop, because all our MySQL indices are prefixed with `shop_id`. With the current iterators, you either iterate over all products and risk timeouts due to inefficient queries, or you iterate over just shops and do the product iteration yourself, leading to a single "iteration" taking a long time due to the number of products.

It would be nice if we supported this directly in the iteration framework, with a compound cursor (something like `[shop_id, product_id]`). The syntax could look something like

```
enumerator_builder.active_record_on_nested_records([Shop.all, Product.all, ProductVariant.all], cursor: cursor)
# Product.all automagically gets scoped to the shop returned by the first iteration
# Does ProductVariant get scoped to Product, or to Product *and* Shop? Probably Product.

# You can also pass a Proc to do your own custom scoping
enumerator_builder.active_record_on_nested_records([Shop.all, ->(shop) { shop.products }], cursor: cursor)
enumerator_builder.active_record_on_nested_records([Shop.all, ->(shop) { shop.products }, -> (shop, product) { product.product_variants }], cursor: cursor)
```

PR #294, opened 2022-11-01
Replace DatabaseCleaner with Active Record to truncate tables

bdewater:
Fixes CI as noted in https://github.com/Shopify/job-iteration/pull/81#issuecomment-1298085113 cc @sambostock

PR #295, opened 2022-11-02
Support nested iteration

fatkodima:
Fixes https://github.com/Shopify/job-iteration/issues/63.
This is a generalized version of https://github.com/Shopify/job-iteration/pull/100, as suggested in https://github.com/Shopify/job-iteration/issues/63#issuecomment-1150021343.

Example usage:
```ruby
class NestedIterationJob < ApplicationJob

  def build_enumerator(cursor:)
    enumerator_builder.nested(
      [
        ->(cursor) { enumerator_builder.active_record_on_records(Shop.all, cursor: cursor) },
        ->(shop, cursor) { enumerator_builder.active_record_on_records(shop.products, cursor: cursor) },
        ->(_shop, product, cursor) { enumerator_builder.active_record_on_batch_relations(product.product_variants, cursor: cursor) }
      ],
      cursor: cursor
    )
  end

  def each_iteration(product_variants_relation)
    # do something
  end
end
```

cc @bdewater

### DIFF
--- a/lib/job-iteration/nested_enumerator.rb
+++ b/lib/job-iteration/nested_enumerator.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module JobIteration
+  # @private
+  class NestedEnumerator
+    def initialize(enums, cursor: nil)
+      unless enums.all?(Proc)
+        raise ArgumentError, "enums must contain only procs/lambdas"
+      end
+
+      if cursor && enums.size != cursor.size
+        raise ArgumentError, "cursor should have one item per enum"
+      end
+
+      @enums = enums
+      @cursor = cursor || Array.new(enums.size)
+    end
+
+    def each(&block)
+      return to_enum unless block_given?
+
+      iterate([], [], 0, &block)
+    end
+
+    private
+
+    def iterate(current_items, current_cursor, index, &block)
+      cursor = @cursor[index]
+      enum = @enums[index].call(*current_items, cursor)
+
+      enum.each do |item, cursor_value|
+        if index == @cursor.size - 1
+          yield item, current_cursor + [cursor_value]
+        else
+          iterate(current_items + [item], current_cursor + [cursor_value], index + 1, &block)
+        end
+      end
+    end
+  end
+end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -42,6 +42,11 @@ end
 ActiveJob::Base.queue_adapter = :iteration_test
 
 class Product < ActiveRecord::Base
+  has_many :comments
+end
+
+class Comment < ActiveRecord::Base
+  belongs_to :product
 end
 
 host = ENV["USING_DEV"] == "1" ? "job-iteration.railgun" : "localhost"
@@ -68,9 +73,16 @@ Sidekiq.configure_client do |config|
   config.redis = { host: host }
 end
 
-ActiveRecord::Base.connection.create_table(Product.table_name, force: true) do |t|
-  t.string(:name)
-  t.timestamps
+ActiveRecord::Schema.define do
+  create_table(:products, force: true) do |t|
+    t.string(:name)
+    t.timestamps
+  end
+
+  create_table(:comments, force: true) do |t|
+    t.string(:content)
+    t.belongs_to(:product)
+  end
 end
 
 module LoggingHelpers
@@ -118,9 +130,16 @@ class IterationUnitTest < ActiveSupport::TestCase
   end
 
   def insert_fixtures
-    10.times do |n|
-      Product.create!(name: "lipstick #{n}")
-    end
+    now = Time.now
+    products = 10.times.map { |n| { name: "lipstick #{n}", created_at: now - n, updated_at: now - n } }
+    Product.insert_all!(products)
+
+    comments = Product.order(:id).limit(3).map.with_index do |product, index|
+      comments_count = index + 1
+      comments_count.times.map { |n| { content: "#{product.name} comment ##{n}", product_id: product.id } }
+    end.flatten
+
+    Comment.insert_all!(comments)
   end
 
   def truncate_fixtures
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### Master (unreleased)
 
+- [295](https://github.com/Shopify/job-iteration/pull/295) - Support nested iteration
 - [241](https://github.com/Shopify/job-iteration/pull/241) - Require Ruby 2.7+, dropping 2.6 support
 - [241](https://github.com/Shopify/job-iteration/pull/241) - Require Rails 6.0+, dropping 5.2 support
 - [240](https://github.com/Shopify/job-iteration/pull/240) - Allow setting inheritable per-job `job_iteration_max_job_runtime`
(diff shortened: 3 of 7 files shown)


fatkodima:
I signed CLA, a re-run is needed.

bdewater:
Thank you for tackling this 🙇 the API looks great!

Once CI runs you'll hit failures against Rails edge, you may want to get ahead of that by borrowing https://github.com/Shopify/job-iteration/pull/294

Mangara:
Would you mind rebasing on the latest `main`? We merged the fix for the Rails edge tests, so CI should be 🟢 after that.

fatkodima:
Rebased.

Mangara:
There are some CI failures 🤔 

```
1) Failure:
JobIteration::NestedEnumeratorTest#test_cursor_can_be_used_to_resume [/home/runner/work/job-iteration/job-iteration/test/unit/nested_enumerator_test.rb:48]:
--- expected
+++ actual
@@ -1 +1 @@
-[#<Comment id: 7, content: "lipstick 0 comment #0", product_id: 1>, [1, 7]]
+[#<Comment id: 2, content: "lipstick 1 comment #0", product_id: 2>, [2, 2]]
```

Is this test flaky? Or did one of the gem bumps break it?

fatkodima:
@Mangara I identified what is the problem. 

Do not know how I overlooked it, but this iterator isn't working correctly when created from an existing cursor when resuming iteration. Each cursor in the chain is independently advanced. For example, if the cursor was `[1, 2]`, the next yielded record will have the cursor `[2, 3]` (instead of `[1, 3]` or `[2, 1]`).

And additionally the test was weak enough to identify it when running in isolation.

The simplest solution is to just store the previous cursor https://github.com/Shopify/job-iteration/blob/b1f818e4f0be1317eddf89bafdd8e6ec9a5a3566/lib/job-iteration/iteration.rb#L185 when the job is interrupted. But the next time the same record will be processed once again. Is it a problem? In ActiveJob/Sidekiq each job already can be processed multiple times because of errors/etc, so people are already encouraged to write idempotent code in jobs.

We can slightly improve the situation with the same record reruns mentioned above, by storing *previous* cursor value for cursors with multiple values, and the *current* value having only one.

test/test_helper.rb
```diff
@@ -118,9 +130,16 @@ class IterationUnitTest < ActiveSupport::TestCase
...
+
+    comments = Product.order(:id).limit(3).map.with_index do |product, index|
+      comments_count = index + 1
+      comments_count.times.map { |n| { content: "#{product.name} comment ##{n}", product_id: product.id } }
+    end.flatten
+
+    Comment.insert_all!(comments)
   end
 
   def truncate_fixtures
```
bdewater, line 145: I think you want to change the line below to `ActiveRecord::Base.connection.truncate(Product.table_name, Comment.table_name)`

bdewater:
> But the next time the same record will be processed once again. Is it a problem?

Is it different than how it is today? I had a bit of trouble parsing the test cases, so I played with these to make sure I understand it correctly (and please correct me if I'm still wrong)
```ruby
test "array cursor can be used to resume" do
  enumerator_builder = EnumeratorBuilder.new(mock)
  integers = [0, 1, 2]

  enum = enumerator_builder.array(integers, cursor: 0)

  assert_equal [1, 2], enum.map { |value, _cursor| value }
end

test "nested array cursor can be used to resume" do
  enumerator_builder = EnumeratorBuilder.new(mock)
  integers = [0, 1, 2]
  strings = [["0a", "0b", "0c"], ["1a", "1b", "1c"], ["2a", "2b", "2c"]]

  enum = enumerator_builder.nested(
    [
      ->(cursor) { enumerator_builder.array(integers, cursor: cursor) },
      ->(integer, cursor) { enumerator_builder.array(strings[integer], cursor: cursor) },
    ],
    cursor: [0, 0]
  )

  assert_equal ["1b", "1c", "2a", "2b", "2c"], enum.map { |value, _cursor| value }
end
```
The second test is correctly specified, but currently fails with:
```
Expected: ["1b", "1c", "2a", "2b", "2c"]
  Actual: ["1b", "1c", "2b", "2c"]
```

fatkodima:
> Is it different than how it is today?

Yes. Currently (`main` branch), resumed iteration will start from the `cursor` + 1 (actually, passed `cursor`s value is a value from the latest previous iteration).

The first test case has `assert_equal [1, 2], enum.map { |value, _cursor| value }`, but with the new behavior of starting from the latest `cursor` (without +1) will be `assert_equal [0, 1, 2], enum.map { |value, _cursor| value }`.

Second test case: since we are starting from `[0, 0]`, the expected next value should be from `[0, 1]`, isn't it? - that's why we need to restart from the previous cursor value because otherwise some values will be incorrectly skipped when doing nested iteration.

Not sure if it is more clear now. Feel free to ask, and I will try rephrase and somehow improve my explanation.

lib/job-iteration/nested_enumerator.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    private
+
+    def iterate(current_items, current_cursor, index, &block)
+      cursor = @cursor[index]
+      enum = @enums[index].call(*current_items, cursor)
+
+      enum.each do |item, cursor_value|
+        if index == @cursor.size - 1
+          yield item, current_cursor + [cursor_value]
```
bdewater, line 33: I've been stepping through the debugger a few times and what stood out to me is `cursor = @cursor[index]`, the initial cursor value stays 'stuck' for subsequent iterations. In my array example this means skipping over `"2a"` but also `"3a"` if you extend the pattern as such:

```ruby
integers = [0, 1, 2, 3]
strings = [["0a", "0b", "0c"], ["1a", "1b", "1c"], ["2a", "2b", "2c"], ["3a", "3b", "3c"]]
```

This seems to do the trick unless I'm missing something (it's getting late here)?
```suggestion
          yield item, current_cursor + [cursor_value]
          @cursor[index] = nil
```

bdewater: I guess what I am not grasping is why the cursor logic throughout this library needs to change to make nesting work. AFAIK its `nil` to start at the beginning of a collection, and a non-nil cursor to resume _after_. With the generic way nested enumerators are set up I don't see why this should change? The individual enumerators shouldn't need to care if it's nested or not, the nested enumerator should feed the correct value to them and handle the nesting level (`index`).

To whiteboard the possibly cursor values for `["0a", "0b", "0c"]` and the item processed in each_iteration:
- cursor: `nil`, item: "0a"
- cursor: `0`, item: "0b"
- cursor: `1`, item: "0c"

and nested with: `[["0a", "0b", "0c"], ["1a", "1b", "1c"], ["2a", "2b", "2c"]]`:
- cursor: `[nil, nil]`, item: "0a"
- cursor: `[nil, 0]`, item: "0b"
- cursor: `[nil, 1]`, item: "0c"
- cursor: `[0, nil]`, item: "1a"
- cursor: `[0, 0]`, item: "1b"
- cursor: `[0, 1]`, item: "1c"
- cursor: `[1, nil]`, item: "2a"

etc. It seems to me resetting the right cursor index to nil at the correct time should make the above possible and make it work without changes to the rest of the library.

Having debugged some more I think it's also possible to refactor out `current_cursor` in the iterate method, it seems redundant with `@cursor` being initialized (from the interrupted cursor or an array of nils) but never updated 🤔 I'll try to get a PR up in the next few days. 

Either way - thanks for taking the time to work on this and answering my questions 🙇‍♂️

PR #306, opened 2022-11-29
Update README.md

milan-chudik-shpfy:
I believe `ApplicationJob` used as parent class requires extra "domain" knowledge, where this class is not found anywhere in `job-iteration` or `activejob`.gems. In fact only `ActiveJob::Base` is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Meet Iteration, an extension for [ActiveJob](https://github.com/rails/rails/tree
 Imagine the following job:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


Mangara:
`ApplicationJob` is created automatically for new Rails projects and the [Active Job guide](https://guides.rubyonrails.org/active_job_basics.html#creating-a-job) uses it as the base class for all jobs. I wonder if using `ActiveJob::Base` wouldn't be more confusing for beginner Rails devs, instead of less?

PR #338, opened 2023-02-14
Use ActiveSupport::Notifications and LogSubscriber for customizing log output

bdewater:
This is the [same mechanism](https://api.rubyonrails.org/classes/ActiveSupport/LogSubscriber.html) Rails uses to output logs from events. This allows for more instrumentation and easier customization of log output. The default log output is completely unchanged.

Since a bunch of events need `cursor_position` to keep the existing logging, I made it a default tag for all events.

Edit: closes https://github.com/Shopify/job-iteration/pull/40 by including it in an updated way.

### DIFF
--- a/lib/job-iteration.rb
+++ b/lib/job-iteration.rb
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
 
+require "active_job"
 require_relative "./job-iteration/version"
 require_relative "./job-iteration/enumerator_builder"
 require_relative "./job-iteration/iteration"
+require_relative "./job-iteration/log_subscriber"
 
 module JobIteration
   IntegrationLoadError = Class.new(StandardError)
@@ -10,6 +12,10 @@ module JobIteration
   INTEGRATIONS = [:resque, :sidekiq]
 
   extend self
+
+  attr_accessor :logger
+
+  self.logger = ActiveJob::Base.logger
 
   # Use this to _always_ interrupt the job after it's been running for more than N seconds.
   # @example
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -142,13 +142,12 @@ module JobIteration
       self.start_time = Time.now.utc
 
       enumerator = nil
-      ActiveSupport::Notifications.instrument("build_enumerator.iteration", iteration_instrumentation_tags) do
+      ActiveSupport::Notifications.instrument("build_enumerator.iteration", instrumentation_tags) do
         enumerator = build_enumerator(*arguments, cursor: cursor_position)
       end
 
       unless enumerator
-        logger.info("[JobIteration::Iteration] `build_enumerator` returned nil. " \
-          "Skipping the job.")
+        ActiveSupport::Notifications.instrument("nil_enumerator.iteration", instrumentation_tags)
         return
       end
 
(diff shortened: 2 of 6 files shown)


bdewater:
Test failure on Ruby head/Rails head, insofar other combinations ran it seemed to succeed.

```
  1) Failure:
ResqueIntegrationTest#test_interrupts_the_job [/home/runner/work/job-iteration/job-iteration/test/integration/integration_behaviour.rb:22]:
Expected: 0
  Actual: nil
```

lib/job-iteration/iteration.rb
```diff
@@ -171,7 +171,7 @@ def interruptible_perform(*arguments)
         reenqueue_iteration_job
       elsif completed
         run_callbacks(:complete)
-        output_interrupt_summary
+        ActiveSupport::Notifications.instrument("completed.iteration", tags)
```
Mangara, line 174: Will the `total_time` tag be accurate here, if we capture it before iterating?

lib/job-iteration/iteration.rb
```diff
@@ -184,10 +184,11 @@ def iterate_with_enumerator(enumerator, arguments)
         # Deferred until 2.0.0
         # assert_valid_cursor!(index)
 
-        record_unit_of_work do
+        ActiveSupport::Notifications.instrument("each_iteration.iteration", {}) do |tags|
           found_record = true
           each_iteration(object_from_enumerator, *arguments)
           self.cursor_position = index
+          tags.replace(instrumentation_tags)
```
Mangara, line 191: Do we really want the updated `cursor_position` here? The notification describes the iteration that just happened, so it makes most sense to me to tag it with the cursor position at the start of that iteration. If we do want to include both, we can give the one after a different name?

lib/job-iteration/log_subscriber.rb
```diff
@@ -0,0 +1,38 @@
...
+module JobIteration
+  class LogSubscriber < ActiveSupport::LogSubscriber
+    def logger
+      JobIteration.logger
+    end
+
+    def nil_enumerator(event)
+      info do
+        "[JobIteration::Iteration] `build_enumerator` returned nil. Skipping the job."
+      end
```
Mangara, line 12: Why pass a block here, instead of the string itself?
```suggestion
      info("[JobIteration::Iteration] `build_enumerator` returned nil. Skipping the job.")
```

Mangara (commented):
I like the approach! Just some small comments.

lib/job-iteration/log_subscriber.rb
Mangara: Ah, right. I was thinking about the overhead of calling into the block when you're logging it, but didn't consider the string interpolation cost when not logging it. I guess it comes down to optimizing for log levels above or below `info`, and how much difference it makes in those cases. Either way, I'm not too bothered - I'm fine keeping the blocks if you think that makes the most sense.

lib/job-iteration/iteration.rb
Mangara: Okay, that makes sense. I think my confusion comes from the fact that when you enqueue the job with `cursor_position: 1`, it doesn't process id 1, but only 2 and up, so the cursor position and item processed don't quite line up. But this is indeed the `cursor_position` of the thing we just processed, so it makes sense to use this as the tag 👍

lib/job-iteration.rb
```diff
@@ -11,6 +13,10 @@ module JobIteration
 
   extend self
 
+  attr_accessor :logger
+
+  self.logger = ActiveJob::Base.logger
```
etiennebarrie, line 18: This does change something. Previously we ended up logging to `Rails.logger`, because when ActiveJob initializes itself in the context of a Rails application, it sets its logger to the Rails one:
https://github.com/rails/rails/blob/v7.0.7.2/activejob/lib/active_job/railtie.rb#L13-L15

Whereas now it will use the default Active Job logger which logs to stdout: https://github.com/rails/rails/blob/v7.0.7.2/activejob/lib/active_job/logging.rb#L11

bdewater: TIL. Thanks for also providing a fix - linking https://github.com/Shopify/job-iteration/pull/427 here in case others run into this problem.

PR #339, opened 2023-02-14
Continue other jobs on Rails edge failures

sambostock:
Rails edge (and Ruby HEAD, but we always combine that with Rails edge) is a moving target out of our control. Therefore, if a failure occurs on a job testing against it, it is useful to let the other jobs continue, as that helps narrow down if the failure is due to a change made there.

I noticed that CI was failing on #81, but couldn't tell if it was isolated to edge versions because the jobs would get cancelled. Sure enough, it looks like something has broken upstream.


sambostock:
CI failures on Ruby HEAD are expected. See steveklabnik/mono_logger#12.

Mangara (approved):

PR #340, opened 2023-02-15
Add cursor next iteration metric

etravassos:


lib/job-iteration/active_record_enumerator.rb
```diff
@@ -39,6 +39,12 @@ def size
 
     private
 
+    def instrument_next_batch(cursor)
+      ActiveSupport::Notifications.instrument("cursor.iteration") do
```
etiennebarrie, line 43:
```suggestion
      ActiveSupport::Notifications.instrument("active_record_cursor.iteration") do
```

etiennebarrie (commented):
There's at least the ActiveRecordBatchEnumerator that would be interesting to instrument as well. But yes it's a good idea!

lib/job-iteration/active_record_enumerator.rb
etravassos: Hey 👋  @etiennebarrie @Mangara 
Pushed the tweak if you could take another look.

PR #346, opened 2023-02-22
Adjust total time for failed jobs

emily827:
This PR ensures that `total_time` is always going to be updated, even when a job fails.

Previously, if a job failed, `total_time` would not include the run-time of the iteration that threw the exception.

**Motivation**
We want to create a metric to keep track of a job's total run-time. This metric is going to be reported whenever a job completes successfully, fails, or is interrupted


solackerman (approved):
makes sense :+1: it looks like some tests are failing though.

emily827:
> makes sense 👍 it looks like some tests are failing though.

I think that's because the main branch is failing 🤔 Those failures don't look related to my changes I think

PR #347, opened 2023-02-23
Only call adjust_total_time once

emily827:
Previously, I introduced a [PR](https://github.com/Shopify/job-iteration/pull/346)  that caused `total_time` to be updated if a job is interrupted, fails, or succeeds.

However, this also introduced a bug where if the job is interrupted, adjust_total_time is actually called twice- once in the `ensure` statement of the `iterate_with_enumerator` method and again in the `reenqueue_iteration_job` method, resulting in incorrect `total_time` since the current iteration's runtime would be double-counted

I added a new test where a job is interrupted, and verified that `total_time` is correct after the interruption.


solackerman (approved):

iamrestrepo (approved):
🎊  Thanks!

PR #348, opened 2023-02-24
Fix a bug in readme

adamzapasnik:
kind of threw me off, because job's args are always passed to `each_iteration`, right?


Mangara (approved):
Good catch!

PR #360, opened 2023-03-26
Fix incorrect cursor position in instrumentation

bdewater:
... when an exception is raised inside each_iteration. Add tests to cover both successful and erroring iterations. Follow-up to https://github.com/Shopify/job-iteration/pull/338.

I snuck in a small rename here, I always found the index - cursor_position difference confusing to read. Cursors can be anything, not just an index. It makes it more clear that an enumerator must yield the cursor position that will be serialized if the job is interrupted.

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -185,15 +185,15 @@ module JobIteration
       found_record = false
       @needs_reenqueue = false
 
-      enumerator.each do |object_from_enumerator, index|
+      enumerator.each do |object_from_enumerator, cursor_from_enumerator|
         # Deferred until 2.0.0
-        # assert_valid_cursor!(index)
+        # assert_valid_cursor!(cursor_from_enumerator)
 
-        ActiveSupport::Notifications.instrument("each_iteration.iteration", {}) do |tags|
+        tags = instrumentation_tags.merge(cursor_position: cursor_from_enumerator)
+        ActiveSupport::Notifications.instrument("each_iteration.iteration", tags) do
           found_record = true
           each_iteration(object_from_enumerator, *arguments)
-          self.cursor_position = index
-          tags.replace(instrumentation_tags)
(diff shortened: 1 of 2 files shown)


lib/job-iteration/iteration.rb
```diff
@@ -185,15 +185,15 @@ def iterate_with_enumerator(enumerator, arguments)
...
       @needs_reenqueue = false
 
-      enumerator.each do |object_from_enumerator, index|
+      enumerator.each do |object_from_enumerator, cursor_position|
         # Deferred until 2.0.0
-        # assert_valid_cursor!(index)
+        # assert_valid_cursor!(cursor_position)
 
-        ActiveSupport::Notifications.instrument("each_iteration.iteration", {}) do |tags|
+        tags = { job_class: self.class.name, cursor_position: cursor_position }
```
Mangara, line 192: Calling `instrumentation_tags` here is useful if we want to add more default tags in the future.
```suggestion
        tags = instrumentation_tags.merge(cursor_position: cursor_position)
```

test/unit/iteration_test.rb
```diff
@@ -414,6 +414,33 @@ def test_total_time_is_updated_for_failed_jobs
     end
   end
 
+  def test_each_iteration_instrumentation
+    notifications = []
+    ActiveSupport::Notifications.monotonic_subscribe("each_iteration.iteration") do |_, _, _, _, tags|
```
Mangara, line 419: Can we use [`ActiveSupport::Notifications.subscribed`](https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html#module-ActiveSupport::Notifications-label-Temporary+Subscriptions) here, so the subscription is removed after the test finishes?

Mangara (commented):
The Ruby 2.7 test failures seem legit.

bdewater:
@Mangara fixed :)

PR #362, opened 2023-03-26
[doc] Document argument passing semantics

julik:
It is not really clear from the documentation how arguments to `perform` are supposed to propagate through the call stack, and if one needs to blind-explore it leads to some rather surprising results. Maybe documenting it could help?..

### DIFF
--- a/guides/argument-semantics.md
+++ b/guides/argument-semantics.md
@@ -23,7 +23,7 @@ class ArglessJob < ActiveJob::Base
   end
 end
 
-To spool the job:
+To enqueue the job:
 
 ```ruby
 ArglessJob.perform_later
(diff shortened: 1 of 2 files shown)


guides/argument-semantics.md
```diff
@@ -0,0 +1,127 @@
...
+  def build_enumerator(cursor:)
+    # ...
+  end
+
+  def each_iteration(single_object_yielded_from_enumerator)
+    # ...
+  end
+end
+
+To spool the job:
```
bdewater, line 26: I'd change this and below to enqueue, [like the Active Job docs](https://guides.rubyonrails.org/active_job_basics.html#enqueue-the-job).

julik:
This will likely address https://github.com/Shopify/job-iteration/issues/220 as well

PR #366, opened 2023-03-29
Yield updated instead of initial nested cursor to persist

bdewater:
Otherwise we reprocess the last successfully processed object again when resuming. In a pathologic case, if a nested job is interrupted every time after a single `each_iteration` completes we'd never progress: the cursor would stay at `[nil, nil]`.

I spelled it out in https://github.com/Shopify/job-iteration/pull/364 😅 but I originally worked on this months before and only realized it now.

Follow-up to https://github.com/Shopify/job-iteration/pull/310


Mangara (approved):
Good catch! I like that we can remove the `dup` from the instrumented tags too.

PR #367, opened 2023-04-02
Infer interruption handler from a job's queue adapter

bdewater:
... and allow Iteration to be used with multiple job backends simultaneously. Closes https://github.com/Shopify/job-iteration/pull/96, it's based on this PR but it looks abandoned.

The `Fallback` adapter mimics previous behaviour if `self.interruption_adapter = -> { false }` was not overridden for Resque or Sidekiq, but to me it seems more correct to raise an error instead of no interruption adapter is found. If the maintainers are open to a breaking change I can make it so.

Removed test_mark_job_worker_as_interrupted since it was testing stubs.

### DIFF
--- a/lib/job-iteration/integrations.rb
+++ b/lib/job-iteration/integrations.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module JobIteration
+  # @api private
+  module Integrations
+    LoadError = Class.new(StandardError)
+
+    extend self
+
+    attr_accessor :registered_integrations
+
+    self.registered_integrations = {}
+
+    autoload :Sidekiq, "job-iteration/integrations/sidekiq"
+    autoload :Resque, "job-iteration/integrations/resque"
+
+    # @api public
+    def register(name, callable)
+      raise ArgumentError, "Interruption adapter must respond to #call" unless callable.respond_to?(:call)
+
+      registered_integrations[name] = callable
+    end
+
+    def load(name)
+      if (callable = registered_integrations[name])
+        callable
+      else
+        begin
+          klass = "#{self}::#{name.camelize}".constantize
+          register(name, klass)
+        rescue NameError
+          raise LoadError, "Could not find integration for '#{name}'"
+        end
+      end
+    end
+  end
+end
--- a/lib/job-iteration/integrations/resque.rb
+++ b/lib/job-iteration/integrations/resque.rb
@@ -4,21 +4,27 @@ require "resque"
 
 module JobIteration
   module Integrations
-    module ResqueIterationExtension # @private
-      def initialize(*) # @private
-        $resque_worker = self
-        super
+    module Resque
+      module IterationExtension
+        def initialize(*)
+          $resque_worker = self
+          super
+        end
+      end
+
+      # @private
+      module ::Resque
+        class Worker
+          # The patch is required in order to call shutdown? on a Resque::Worker instance
+          prepend(IterationExtension)
+        end
+      end
+
+      class << self
+        def call
+          $resque_worker.try!(:shutdown?)
+        end
       end
     end
-
-    # @private
-    module ::Resque
-      class Worker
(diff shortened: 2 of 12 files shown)


Mangara (commented):
Can we keep `JobIteration. interruption_adapter` around for queue adapters other than Sidekiq or Resque? Or should we give them a way to hook into the discovery? One tricky bit there is that `queue_adapter_name` can return `class` if you do something like `config.active_job.queue_adapter = OtherQueueAdapter`.

> If the maintainers are open to a breaking change I can make it so.

Removing a configuration option is already a breaking change 🙂 Even without that, merged unreleased changes include dropping support for older Ruby versions and such, so we're already looking at a major release I think. I'm fine wth raising if no interruption configuration is found.

lib/job-iteration/integrations.rb
```diff
@@ -0,0 +1,26 @@
...
+    LoadError = Class.new(StandardError)
+
+    autoload :Sidekiq, "job-iteration/integrations/sidekiq"
+    autoload :Resque, "job-iteration/integrations/resque"
+
+    extend self
+
+    def load(job)
+      camelized_name = job.class.queue_adapter_name.camelize
+      object = const_get(camelized_name)
```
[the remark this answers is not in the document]
bdewater: > Interesting. Recommending that users monkey patch your library feels off to me, there must be better ways to provide that configuration. How do you feel about a lookup table from `queue_adapter_name` to interruption adapter that users can add to?

Maybe I was pattern matching too much on Active Job/GoodJob, your suggestion makes sense to me.

> That way, queue adapters that trigger the Rails bug can just register for `"class"` and we don't have to work around it here. (I [fixed the bug](https://github.com/rails/rails/pull/47995), btw 🙂, although that only helps for Rails 7.1 and on)

Awesome 🙌 it would be useful to backport to 7.0 IMO, for those of us not living on Rails main 😛 I don't think it returning `"module"` is useful/expected.

plasticine:
Thanks for picking this up @bdewater 👏 priorities changed a bit and I never got back to #96, so thanks for pushing it forward!

bdewater:
Gentle nudge @Mangara :)

lib/job-iteration/integrations/resque.rb
```diff
@@ -4,21 +4,27 @@
...
-    module ::Resque
-      class Worker
-        # The patch is required in order to call shutdown? on a Resque::Worker instance
-        prepend(ResqueIterationExtension)
+      # @private
+      module ::Resque
+        class Worker
+          # The patch is required in order to call shutdown? on a Resque::Worker instance
+          prepend(IterationExtension)
+        end
```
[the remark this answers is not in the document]
bdewater: This breaks the YARD CI step: https://github.com/Shopify/job-iteration/actions/runs/5982086854/job/16230774487?pr=367

Mangara (approved):
Thanks for all your hard work on this, and apologies for the long delay!

lib/job-iteration/integrations.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def load(name)
+      if (callable = registered_integrations[name])
+        callable
+      else
+        begin
+          klass = "#{self}::#{name.camelize}".constantize
+          register(name, klass)
+        rescue NameError
+          raise LoadError, "Could not find integration for '#{name}'"
```
lavoiesl, line 32: This should register a dummy handler, not raise.
Otherwise, this results in an exception for any queue adapter that isn't recognized

lavoiesl: But this breaks the users (our team uses active_qless (yes, I know, it's unmaintained, we're moving away from it))

How about this:
- Add a graceful fallback, preserving the current behaviour, but with a deprecated notice
- Release a patch version on the current minor version
- Plan to remove the graceful fallback in the next major version

lavoiesl: Alright then 👍🏼

PR #418, opened 2023-08-18
Return `nil` from `Iteration#perform`

sambostock:
`perform` should be treated as `void`, with consumers not coupling to its return value.

Explicitly returning `nil` ensures the return value does not change if we happen to change the implementation.

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -125,6 +125,8 @@ module JobIteration
 
     def perform(*params) # @private
       interruptible_perform(*params)
+
+      nil
     end
 
     def retry_job(*, **)
--- a/test/unit/iteration_test.rb
+++ b/test/unit/iteration_test.rb
@@ -485,6 +485,11 @@ class JobIterationTest < IterationUnitTest
     assert_equal(expected, events)
   end
 
+  def test_perform_returns_nil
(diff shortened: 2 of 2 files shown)


lib/job-iteration/iteration.rb
```diff
@@ -178,6 +178,8 @@ def interruptible_perform(*arguments)
           instrumentation_tags.merge(times_interrupted: times_interrupted, total_time: total_time),
         )
       end
+
+      nil
```
sambostock, line 182: I put this here because `perform` just delegates to this method, but maybe it would make more sense to put it there?


```diff
 def perform(*params) # @private
   interruptible_perform(*params)
+  
+  nil
 end
 
 def interruptible_perform(*arguments)
   # ...
-  
-  nil
 end
```

Not sure...

Mangara: We're making this change because we care about the return value of `perform`, so it makes sense to me to put it there, instead of here. Not a big deal, though 🤷

Mangara (approved):

lib/job-iteration/iteration.rb
sambostock: Yeah, that makes sense. I've updated it.

Mangara (approved):

PR #419, opened 2023-08-18
Extract separate CI linting job

sambostock:
Because Rubocop allows us to specify a `TargetRubyVersion`, we don't need to run the linting step across every Ruby version – we can simply run it with the latest, while setting the `TargetRubyVersion` to the oldest.[^1]

[^1]: Tests are added to ensure these values are kept in sync.


Mangara (approved):
I love it!

PR #431, opened 2023-09-12
Support AR Enumerators for models with composite primary keys using :id

adrianna-chang-shopify:
We need to use `#id_value` instead of `#id` for models with composite primary keys when accessing the record value, because `#id` gives us back the primary key value when we want the `:id` column value.


nvasilevski (approved):
👍

Mangara:
Thanks for fixing this! Could you add it to the change log as unreleased?

adrianna-chang-shopify:
Yes, thanks for the nudge @Mangara !

Issue #279, opened 2022-09-14
How to handle retries for exceptions?


juanmrad:
I have a question regarding handling retries for exceptions.

From my understanding it seems like if an exception happens the whole job would be retried from the beginning. Is it possible to adjust so it only retries from where it left of?

Examples. 
Job is inserting data on each iteration and one of them fails due to DB/Connection error. I want it to retry the job but only from that iteration forward not from the beginning. Is the cursor persisted on the retry and use that?

or on a similar vein I'm sending data to an API and get rate limited and need to retry with a backoff, but only from where it was. will Job iterator handle this or is there an easy way to handle this?

Mangara:
> From my understanding it seems like if an exception happens the whole job would be retried from the beginning. Is it possible to adjust so it only retries from where it left of?

If iterations 1 through 4 are successful, and iteration 5 fails, the cursor is persisted and the retry will start at iteration 5.

The [FAQ](https://github.com/Shopify/job-iteration#faq) states

> **What happens with retries?** An interruption of a job does not count as a retry. The iteration of job that caused the job to fail will be retried and progress will continue from there on.

Is there a way we can clarify that?

juanmrad:
@Mangara Thanks for the clarification.

I wasn't sure if after each iteration the cursor is committed/set on the job such that when going on iteration 5 and it fails the cursor will be available to ensure that the job is not performed from the beginning.

The answer made it seem like interruptions are not retries so the job continues from where it was, but does not clarify that any exception that is retried during an iteration will keep the cursor of that iteration to be retried and continued from there.

juanmrad:
@Mangara Hi, sorry for the extra ping, wanted to confirm my assumption was correct that if the job is not interrupted by a shutdown and it is iterating the cursor is part of the job so if a failure happens on an iteration that causes a retry it will continue from that iteration over the beginning.

Mangara:
> if a failure happens on an iteration that causes a retry it will continue from that iteration

This is correct.

PR #433, opened 2023-10-18
Explain exceptions inside each_iteration

jmeinerz:
It wasn't entirely clear to us what would happen when the code blows up during a given iteration. We ran a quick experiment and this PR documents what we found. We think this may be useful to others as well.

This hopefully brings clarity to https://github.com/Shopify/job-iteration/issues/279

### DIFF
--- a/guides/iteration-how-it-works.md
+++ b/guides/iteration-how-it-works.md
@@ -21,6 +21,12 @@ SELECT  `products`.* FROM `products` ORDER BY products.id LIMIT 100
 SELECT  `products`.* FROM `products` WHERE (products.id > 2) ORDER BY products.id LIMIT 100
 ```
 
+## Exceptions inside `each_iteration`
+
+When an unrescued exception happens inside the `each_iteration` block, the job will stop and re-enqueue itself with the last successful cursor. This means that the iteration that failed will be retried with the same parameters and the cursor will only move if that iteration succeeds. This behaviour may be enough for intermittent errors, such as network connection failures, but if your execution is deterministic and you have an error, subsequent iterations will never run.
+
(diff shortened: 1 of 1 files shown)


guides/iteration-how-it-works.md
```diff
@@ -20,6 +20,11 @@ SELECT  `products`.* FROM `products` ORDER BY products.id LIMIT 100
 ```sql
 SELECT  `products`.* FROM `products` WHERE (products.id > 2) ORDER BY products.id LIMIT 100
 ```
+## Exceptions inside `each_iteration`
+
+When an unrescued exception happens inside the `each_iteration` block, the job will stop and re-enqueue itself with the last successful cursor. This means that the iteration that failed will be retried with the same parameters and the cursor will only move if that iteration succeeds. This behaviour may be enough for intermittent errors, such as network connection failures, but if your execution is deterministic and you have an error, subsequent iterations will never run.
+
+In other words, if you are trying to process 100 records but you have a problem with the 61st, only the first 60 will be processed.
```
Mangara, line 27: For additional clarity:
```suggestion
In other words, if you are trying to process 100 records but the job consistently fails on the 61st, only the first 60 will be processed and the job will try to process the 61st record until retries are exhausted.
```

guides/iteration-how-it-works.md
```diff
@@ -20,6 +20,11 @@ SELECT  `products`.* FROM `products` ORDER BY products.id LIMIT 100
 ```sql
 SELECT  `products`.* FROM `products` WHERE (products.id > 2) ORDER BY products.id LIMIT 100
 ```
+## Exceptions inside `each_iteration`
```
Mangara, line 23: Nit: I'd prefer a bank line to separate this:

```suggestion

## Exceptions inside `each_iteration`
```

Mangara (approved):
Thanks for adding this!

PR #437, opened 2023-11-15
Use min max job runtime between class & global

sambostock:
The existing approach set `JobIteration.max_job_runtime` as the default value of the `job_iteration_max_job_runtime` `class_attribute`. However, this approach would close around the value at the time the `Iteration` module was included in the host `class`.

This means that if a job class is defined before the host application sets `max_job_runtime`, the setting will not be honoured. For example, if using the `maintenance_tasks` gem, which defines a `TaskJob`, the default would be read when the gem is loaded, prior to initializers running, which is where the host application typically sets the global `max_job_runtime`.

This commit changes the implementation in the following ways:

- We no longer restrict the ability to increase the maximum in child classes. There were ways to get around the restriction, and there may be legitimate scenarios in which an application needs to allow an job to exceed its parent class' maximum job runtime, so we can simplify the implementation by removing the constraint.
- We now take the minimum between the class attribute and the global maximum, which works around the default race, while still ensuring the application can enforce a maximum across all jobs, for stability.

This is an alternative implementation to #436, with the goal of being simpler.

---

Closes #436

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -47,30 +47,9 @@ module JobIteration
 
       class_attribute(
         :job_iteration_max_job_runtime,
-        instance_writer: false,
+        instance_accessor: false,
         instance_predicate: false,
-        default: JobIteration.max_job_runtime,
       )
-
-      singleton_class.prepend(PrependedClassMethods)
-    end
-
-    module PrependedClassMethods
-      def job_iteration_max_job_runtime=(new)
-        existing = job_iteration_max_job_runtime
-
-        if existing && (!new || new > existing)
-          existing_label = existing.inspect
-          new_label = new ? new.inspect : "#{new.inspect} (no limit)"
-          raise(
-            ArgumentError,
-            "job_iteration_max_job_runtime may only decrease; " \
-              "#{self} tried to increase it from #{existing_label} to #{new_label}",
-          )
-        end
-
-        super
-      end
     end
 
     module ClassMethods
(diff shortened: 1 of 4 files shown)


Mangara (approved):
I like this approach much better.

lib/job-iteration/iteration.rb
```diff
@@ -291,13 +270,22 @@ def instrumentation_tags
...
     end
 
+    def job_iteration_max_job_runtime
+      global_max = JobIteration.max_job_runtime
+      class_max = self.class.job_iteration_max_job_runtime
+
+      return global_max unless class_max
+      return class_max unless global_max
+
+      [global_max, class_max].min
```
sambostock, line 286: @Mangara I benchmarked this compared to `compact`, `compact!`, and nested conditionals. This is faster than both `compact` versions, and nearly identical to nested conditionals, so I changed it from our `compact!` version and went with this as I figure it's pretty clear, faster, and avoids allocating arrays entirely.

lib/job-iteration/iteration.rb
```diff
@@ -47,30 +47,9 @@ def inspected_cursor
 
       class_attribute(
         :job_iteration_max_job_runtime,
-        instance_writer: false,
+        instance_accessor: false,
```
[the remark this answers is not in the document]
adrianna-chang-shopify: I thought that was inside the `ClassMethods` scope, my bad 🤦‍♀️

adrianna-chang-shopify (approved):
🎉

lib/job-iteration/iteration.rb
Mangara: Yep, that makes sense to me! I don't think performance is super critical here, as most iterations will contain some external calls (DB queries mostly) that will dominate the running time. But this is still very readable, so if we can get both performance and readability I won't say no 😄

PR #438, opened 2023-11-15
Refactor gemfiles

sambostock:
We have several gemfiles in which we have specific version requirements for `activejob` and `activerecord`, for use in our CI matrix.

The current approach breaks on Ruby HEAD because it is detected as specifying multiple versions/sources for those gems. Instead, this refactors them to use an instance variable approach, similar to that used in Shopify/maintenance_tasks.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,17 @@ gemspec
 gem "sidekiq"
 gem "resque"
 
+if defined?(@rails_gems_requirements) && @rails_gems_requirements
+  # We avoid the `gem "..."` syntax here so Dependabot doesn't try to update these gems.
+  [
+    "activejob",
+    "activerecord",
+  ].each { |name| gem name, @rails_gems_requirements }
+else
+  # gem "activejob" # Set in gemspec
+  gem "activerecord"
+end
+
 gem "mysql2", github: "brianmario/mysql2"
 gem "globalid"
 gem "i18n"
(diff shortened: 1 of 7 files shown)


Gemfile
```diff
@@ -11,6 +11,17 @@ gemspec
 gem "sidekiq"
 gem "resque"
 
+if defined?(@rails_gems_requirements) && @rails_gems_requirements
+  # We avoid the `gem "..."` syntax here so Dependabot doesn't try to update these gems.
```
sambostock, line 15: Copied from https://github.com/Shopify/maintenance_tasks/blob/15e43906c0c705721667ee96d55b8220a514a184/Gemfile#L13

sambostock:
Looks like this approach [works on latest Ruby HEAD](https://github.com/Shopify/job-iteration/actions/runs/6882750277/job/18721903933?pr=438), as opposed to the previous approach which [no longer works as of this writing](https://github.com/Shopify/job-iteration/actions/runs/6880158446/job/18713744117?pr=437).

Mangara (approved):
I don't love the dependabot work-around, but it's better than broken CI.

sambostock:
Yeah, it's weird. Shopify/maintenance_tasks#496 has context on how and why they got to their

```ruby
rails = "rails"
gem rails, @rails_gem_requirement
```

approach, which I adapted.

adrianna-chang-shopify (approved):

PR #442, opened 2023-11-21
Cleanup tests

sambostock:
This makes a number of changes to the test code to:

- simplify some things
- get rid of the noise in the logs

Those who have looked at the test logs previously will know how much nicer this is now:

> ```
> Running bundle exec rake test
> Run options: --seed 11084
> 
> # Running:
> 
> ............S.......S..........SS.....SSS.S...............................................S.S......................S.S.........................
> 
> Finished in 24.729132s, 5.7827 runs/s, 14.8004 assertions/s.
> 
> 143 runs, 366 assertions, 0 failures, 0 errors, 12 skips
> 
> You have skipped tests. Run with --verbose for details.
> ```

The changes have been broken down by commit. The only thing I'm unsure of is if we should have a way to easily disable all the silencing (e.g. skip silencing if `ENV.key?("DEBUG")`).


Mangara (approved):
Nice work! I really appreciate the detailed commit messages.

PR #443, opened 2023-12-21
Sidekiq callback instead of internals for graceful shutdown

bdewater:
Happens at the same time as the old code: https://github.com/sidekiq/sidekiq/blob/73c150d0430a8394cadb5cd49218895b113613a0/lib/sidekiq/launcher.rb#L47-L54 but this is less brittle. The [docs](https://github.com/sidekiq/sidekiq/wiki/Deployment#events) even give a Iteration-like example in Sidekiq Pro to stop on the quiet signal.

`Sidekiq::CLI` is only defined when Sidekiq runs as a standalone process. Since 7.0 it can also run [embedded in the web server process](https://github.com/sidekiq/sidekiq/wiki/Embedding) as `Sidekiq::Embedded`, and using this hook allows Iteration to be compatible with mode of operation. The same events are fired here: https://github.com/sidekiq/sidekiq/blob/73c150d0430a8394cadb5cd49218895b113613a0/lib/sidekiq/embedded.rb#L24-L30

The callback itself exists since Sidekiq 3.0: https://github.com/sidekiq/sidekiq/commit/3d117e60517cd2bc98b760aafc1b38780afb4d0c

### DIFF
--- a/lib/job-iteration/integrations/sidekiq.rb
+++ b/lib/job-iteration/integrations/sidekiq.rb
@@ -4,11 +4,21 @@ require "sidekiq"
 
 module JobIteration
   module Integrations # @private
-    JobIteration.interruption_adapter = -> do
-      if defined?(Sidekiq::CLI) && Sidekiq::CLI.instance
-        Sidekiq::CLI.instance.launcher.stopping?
-      else
-        false
+    module Sidekiq
+      class << self
+        attr_accessor :stopping
+
+        def call
+          stopping
+        end
+      end
+    end
+
+    JobIteration.interruption_adapter = JobIteration::Integrations::Sidekiq
+
+    ::Sidekiq.configure_server do |config|
+      config.on(:quiet) do
+        JobIteration::Integrations::Sidekiq.stopping = true
       end
     end
   end


lib/job-iteration/integrations/sidekiq.rb
```diff
@@ -4,11 +4,19 @@
...
-      if defined?(Sidekiq::CLI) && Sidekiq::CLI.instance
-        Sidekiq::CLI.instance.launcher.stopping?
-      else
-        false
+    module Sidekiq
+      extend self
+
+      attr_accessor :stopping
+
+      alias_method :call, :stopping
```
Mangara, line 12: I'd prefer to be a little less fancy here, and only define the things we need on the singleton class, not instances as well.

```suggestion
      class << self
        attr_accessor :stopping

        def call
          stopping
        end
      end
```

Mangara (approved):
Great call! I see that [Sidekiq 3.0](https://rubygems.org/gems/sidekiq/versions/3.0.0) was released in 2014, so dropping support for older versions is not a concern.

lib/job-iteration/integrations/sidekiq.rb
bdewater-thatch: Fixed!

PR #445, opened 2024-01-02
Suggestion: Add around iterate callback

gustavompo:
There has been some circumstances that it would be great to append a callback to job iterations. More noticeably, that adds extensibility to build some generic handlers, such as metrics collection and logging.
At shopify, a particular use case that'll be especially useful is to deal with the canonical logger and long iterating jobs.

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -42,6 +42,7 @@ module JobIteration
 
     included do |_base|
       define_callbacks :start
+      define_callbacks :iterate
       define_callbacks :shutdown
       define_callbacks :complete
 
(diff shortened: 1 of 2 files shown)


andrewn617 (approved):

bdewater:
Since you mention "metrics collection and logging" you might be interested in  https://github.com/Shopify/job-iteration/pull/338 which added the `each_iteration.iteration` [`ActiveSupport::Notifications`](https://edgeapi.rubyonrails.org/classes/ActiveSupport/Notifications.html) event to subscribe to specifically for these kinds of things.

MatthewRBruce (approved):

PR #446, opened 2024-01-03
Add while  enumerator

angeloocana:
Follow up for: 
- https://github.com/Shopify/shopify/pull/470718

The original implementation of that PR was querying some records in batches using the `enumerator_builder.active_record_on_batches`, collecting the ids for iteration and then deleting the records. The problem is that we don't really need to query and load all the records into memory just to grab the ids to deleted them. We can just run a super fast query like `.exists?`, and then run the delete command in batches while there are records.

This PR adds a new `enumerator_builder.while`, where any code block can be based and the iteration will be allowed when the returned value is true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,25 @@ class NestedIterationJob < ApplicationJob
 end
 ```
 
+
+```ruby
+class DeleteXyzJob < ApplicationJob
+  include JobIteration::Iteration
+
+  def build_enumerator(params)
+    enumerator_builder.while { query_model_xyz(params).exists? }
+  end
+
+  def each_iteration(count, params)
+    query_model_xyz(params).limit(1000).delete_all
+  end
+
+  def query_model_xyz(params)
+    Xyz.where(owner_id: params[:owner_id])
+  end
+end
+```
+
 Iteration hooks into Sidekiq and Resque out of the box to support graceful interruption. No extra configuration is required.
 
 ## Guides
--- a/lib/job-iteration/enumerator_builder.rb
+++ b/lib/job-iteration/enumerator_builder.rb
@@ -181,6 +181,29 @@ module JobIteration
       NestedEnumerator.new(enums, cursor: cursor).each
     end
 
+    # Builds Enumerator for while loop iteration.
+    #
+    # @yield Condition to be evaluated before each iteration, the iteration is allowed if a true value is returned.
+    #
+    # @example
+    #   def build_enumerator(params)
+    #     enumerator_builder.while { query_model_xyz(params).exists? }
+    #   end
+    #
+    #   def each_iteration(_, params)
+    #     query_model_xyz(params).limit(1000).delete_all
+    #   end
+    #
+    #   def query_model_xyz(params)
+    #     Xyz.where(owner_id: params[:owner_id])
+    #   end
+    #
+    def build_while_enumerator(&condition)
+      Enumerator.new do |yielder|
+        yielder << nil while condition.call
+      end
+    end
+
     alias_method :once, :build_once_enumerator
     alias_method :times, :build_times_enumerator
     alias_method :array, :build_array_enumerator
@@ -190,6 +213,7 @@ module JobIteration
     alias_method :throttle, :build_throttle_enumerator
     alias_method :csv, :build_csv_enumerator
     alias_method :nested, :build_nested_enumerator
+    alias_method :while, :build_while_enumerator
 
     private
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - [437](https://github.com/Shopify/job-iteration/pull/437) - Use minimum between per-class `job_iteration_max_job_runtime` and `JobIteration.max_job_runtime`, instead of enforcing only setting decreasing values.
   Because it is possible to change the global or parent values after setting the value on a class, it is not possible to truly enforce the decreasing value constraint. Instead, we now use the minimum between the global value and per-class value. This is considered a non-breaking change, as it should not break any **existing** code, it only removes the constraint on new classes.
+- [446](https://github.com/Shopify/job-iteration/pull/446) - Support while loop enumerator
 
 ### Bug fixes
 
--- a/test/unit/enumerator_builder_test.rb
+++ b/test/unit/enumerator_builder_test.rb
@@ -74,6 +74,15 @@ module JobIteration
       )
     end
 
+    test_builder_method(:build_while_enumerator) do
+      count = 0
+      enum = enumerator_builder(wraps: 0).build_while_enumerator do
+        count < 3 ? (count += 1) : false
+      end
+
+      assert_equal [nil, nil, nil], enum.to_a
+    end
+
     # checks that all the non-alias methods were tested
     raise "methods not tested: #{methods.inspect}" unless methods.empty?
 


angeloocana:
> Hello @angeloocana, can't you use active_record_on_batch_relations in order to avoid the records from being eager loaded?

Hey @rporrasluc ! Thanks for the suggestion, here are the results using the code from the [original PR](https://github.com/Shopify/shopify/pull/470718/files#diff-99fb3ee0568f5833cab69abbf2a58e2a4052de67d2c35eaadff9911095f7f864):

## Testing

### Queries for `active_record_on_batches`:
1) Select all record columns:
```sql
SELECT `delivery_promise_sku_settings`.* FROM `delivery_promise_sku_settings` WHERE `delivery_promise_sku_settings`.`shop_id` = 26371970 AND `delivery_promise_sku_settings`.`delivery_promise_provider_id` = 80485628 ORDER BY delivery_promise_sku_settings.id ASC LIMIT 100
```

2) Delete records using IDs returned from the previous query:
```sql
DELETE FROM `delivery_promise_sku_settings` WHERE `delivery_promise_sku_settings`.`shop_id` = 26371970 AND `delivery_promise_sku_settings`.`id` IN (585506915, 585506916, 585506917)
```

A new iteration is not triggered because the job is aware that there is no next page for the cursor.

### Queries for `active_record_on_batch_relations`:
1) Select record ids:
```sql
SELECT `delivery_promise_sku_settings`.`id` FROM `delivery_promise_sku_settings` WHERE `delivery_promise_sku_settings`.`shop_id` = 26371970 AND `delivery_promise_sku_settings`.`delivery_promise_provider_id` = 80485629 ORDER BY delivery_promise_sku_settings.id LIMIT 100
```

2) Select records again, this time returning id and shop_id using the ids from the previous query: 
```sql
SELECT `delivery_promise_sku_settings`.`id`, `delivery_promise_sku_settings`.`shop_id` FROM `delivery_promise_sku_settings` WHERE `delivery_promise_sku_settings`.`shop_id` = 26371970 AND `delivery_promise_sku_settings`.`delivery_promise_provider_id` = 80485629 AND `delivery_promise_sku_settings`.`id` IN (585506918, 585506919, 585506920) ORDER BY delivery_promise_sku_settings.id
```

3) Select records again, this time returning only the id for the given ids:
```sql
SELECT `delivery_promise_sku_settings`.`id` FROM `delivery_promise_sku_settings` WHERE `delivery_promise_sku_settings`.`id` IN (585506918, 585506919, 585506920) AND `delivery_promise_sku_settings`.`shop_id` = 26371970
```

4) Delete records using IDs from the previous query:
```sql
DELETE FROM `delivery_promise_sku_settings` WHERE `delivery_promise_sku_settings`.`shop_id` = 26371970 AND `delivery_promise_sku_settings`.`id` IN (585506918, 585506919, 585506920)
```

5) On the next iteration run query `1)` again to check if there are remaining records.

### Queries for new `while` proposed solution:

1) Select if there are records to delete (only one or no line is returned):
```sql
SELECT 1 AS one FROM `delivery_promise_sku_settings` WHERE `delivery_promise_sku_settings`.`shop_id` = 26371970 AND `delivery_promise_sku_settings`.`delivery_promise_provider_id` = 80485632 LIMIT 1
```

2) Delete records using the same filter as the previous query to check if records exist:
```sql
DELETE FROM `delivery_promise_sku_settings` WHERE `delivery_promise_sku_settings`.`shop_id` = 26371970 AND `delivery_promise_sku_settings`.`delivery_promise_provider_id` = 80485633 LIMIT 100
```

3) On the next iteration run query `1)` again to check if there are remaining records

## Considerations

Using `active_record_on_batch_relations` made the problem worse, instead of 2 queries, we now have 5 queries.

I really don't see the benefit of querying for the data when we just need to delete it. The job was created because we expect thousands to millions of rows to be deleted. Why do we need to load the data? Maybe I'm missing something...

rporrasluc:
Hello @angeloocana can you share the code where you use `active_record_on_batch_relations`?

nvasilevski:
Ah, well, the original job uses `pluck(:id)` on an Array and not a relation since it uses `active_record_on_batches`(not relations) so pluck itself shouldn't cause a query.
But anyway, what we should be comparing is:
2 queries (`SELECT` ids for the batch, `DELETE` by ids) in `active_record_on_batch_relations` case
2 queries (`SELECT 1` to check if exists, `DELETE` with a limit)

And in this case the benefit is not that clear. Specifically I wonder if deleting by a primary key will perform better due to database doing less row locks. I'm not super familiar with this area but can imagine that `DELETE from 'records' where tenant_id=1 LIMIT 100` may end up performing more locks on the db-level than `DELETE from 'records' where id in (1,2,3, ...100)`

README.md
```diff
@@ -151,6 +151,25 @@ class NestedIterationJob < ApplicationJob
 end
 ```
 
+
+```ruby
+class DeleteXyzJob < ApplicationJob
+  include JobIteration::Iteration
+
+  def build_enumerator(params)
+    enumerator_builder.while query_model_xyz(params).exists?
```
nvasilevski, line 160: While this is a valid ruby syntax, is is an intended one? I believe it was supposed to be a block
```suggestion
    enumerator_builder.while { query_model_xyz(params).exists? }
```

lib/job-iteration/enumerator_builder.rb
```diff
@@ -181,6 +181,30 @@ def build_nested_enumerator(enums, cursor:)
...
+    #   end
+    #
+    #   def query_model_xyz(params)
+    #     Xyz.where(owner_id: params[:owner_id])
+    #   end
+    #
+    def build_while_enumerator(&condition)
+      count = 0
+      Enumerator.new do |yielder|
+        yielder << count += 1 while condition.call
```
nvasilevski, line 204: What would be a use-case for yielding `count`? It will reset to `0` between interruptions and unless there was no interruptions I don't think this number will be useful in any way. 
There should be a way to start counter from the `cursor` position so at least this way `count` will represent the number of times we iterated regardless of whether the job was interrupted or not. 

But perhaps it would be even better to avoid yielding anything until we have a solid use-case for that

angeloocana: I agree 100% , I wasn't sure about what to yield to the job. I'll remove the counter.

adrianna-chang-shopify:
@angeloocana sorry I'm jumping in a little late here, but that second query you've highlighted is specific to code in our private repo. I can elaborate further on the related PR, but in the context of job-iteration, we'd see the same number of queries for the batch relation enumerator as for the while enumerator introduced here (what @nvasilevski described [here](https://github.com/Shopify/job-iteration/pull/446#issuecomment-1881125569)).

Perhaps we could run a benchmark to compare the performance of the two?

@rporrasluc regarding
> Building the relation won't perform any query, we can call delete_all on the relation and it will trigger a DELETE with conditions and a limit. We do not need to pluck ids or check if the relation EXISTS unless I am missing something very obvious.

We do pluck the ids here:
https://github.com/Shopify/job-iteration/blob/f3a89f8f004223322bb2123ce70da1ceffcb4683/lib/job-iteration/active_record_batch_enumerator.rb#L74-L86

This is so that we can build a relation that is already filtered by the primary key values, and yield it to `#each_iteration`:
https://github.com/Shopify/job-iteration/blob/f3a89f8f004223322bb2123ce70da1ceffcb4683/lib/job-iteration/active_record_batch_enumerator.rb#L71

angeloocana:
@adrianna-chang-shopify Thanks for asking for a benchmark, that is what we really needed!

Here are the astonishing results:

### Batches of 1000

Rows | while | active_record_on_batch_relations
-- | -- | --
10 | 0.493995923s | 0.524343613s
100 | 0.558978765s | 0.565930096s
1000 | 0.540043806s | 0.543253054s
10000 | 0.938633587s | 1.368673145s
100000 | 27.29757184s | 16.55144373s
1000000 | 120.7427221s | 89.3870157s

Both solutions have a similar performance for a few thousand rows, with the while solution being a little bit faster. But when the number of records is higher than 100k the `active_record_on_batch_relations` is almost twice as fast 🤯🤯🤯

### Batches of 100

Rows | while | active_record_on_batch_relations
-- | -- | --
100000 | 120.6691102s | 49.36938663s

I also made one test using batches of 100 items (value from the original PR) and it takes 3 to 5 times longer than 1k batches on both solutions.

## Average individual query time

Using batches of 1000

### While
1. Exists? ~160ms
2. Delete all ~300ms

### active_record_on_batch_relations
1. Pluck on id (table to delete data) ~15ms
2. Shop Pluck (shops table) ~0.8ms
3. ShopRoute Pluck (shops table) ~0.8ms
4. Pluck on id and shop_id (table to delete data) ~20ms
5. Delete all ~15ms

## Conclusion

The solution using `active_record_on_batch_relations` performs more queries and transfer more data but it runs faster because it uses the database indexes better.

#### Example queries

Exists for the `while` solution:
```
SELECT  1 AS one FROM `xyz` WHERE `xyz`.`shop_id` = 26371970 AND `xyz`.`owner_id` = 80485674 
LIMIT 1
```

Pluck on id for `active_record_on_batch_relations`
```
SELECT `xyz`.`id` FROM `xyz` WHERE `xyz`.`shop_id` = 26371970 AND `xyz`.`owner_id` = 80485673 AND (xyz.id > '598770502') ORDER BY xyz.id 
LIMIT 1000
```

Both queries filter the data using the same columns but the `active_record_on_batch_relations` also filters by the id value from the last cursor with an order by. I think this makes the database knows exactly what part of the index to use making the query a lot faster.

nvasilevski:
@angeloocana thanks for the investigation! To be honest I'm a little surprised that `exists?` which is (`SELECT 1`) takes 160ms, I would expect it to be much faster as long as there is an index on `shop_id, owner_id`. But I guess if the 300ms for `DELETE` measurement is correct it won't matter since this will still make the `while` approach slower

To be honest I wasn't expecting these results and was mainly leaning towards `active_record_on_batch_relations` because it's a well tested approach that unlikely will bring any surprises while being unnoticeable slower. I was also expecting deletion by the primary key to be somewhat healthier for the platform but it looks like it end up being significantly faster.

Anyway, I wanted to thank you for your proposal as I find the solution very clean and even though our current use-case may not need a new enumerator I can totally see how in the future someone may need "an enumerator that allows the job to run in an interruptible manner until certain condition is met" so I wouldn't be surprised if this PR gets re-opened or implementation will be borrowed for a similar proposal.

Great work!

Issue #309, opened 2022-12-06
Sidekiq's retry mechanism doesn't persist cursor value


adamzapasnik:
versions:
```
rails: 7.0.4
sidekiq: 6.5.8
job-iteration: 1.3.6
```

I'm using activejob ofc. I was surprised when one of the jobs started running from the beginning after a failure and a retry. Found this topic https://github.com/Shopify/job-iteration/issues/279 and figured out this is a bug. 

Looking at the test suite I can see that `retry_on` is covered, but Sidekiq doesn't use `serialize` instead they have custom logic https://github.com/mperham/sidekiq/blob/6-x/lib/sidekiq/job_retry.rb#L140 to serialize a retry job.

There is `sidekiq-iteration` gem that works directly with sidekiq jobs and it seems they monkey patch Sidekiq's `JobRetry` class. https://github.com/fatkodima/sidekiq-iteration/blob/master/lib/sidekiq_iteration/job_retry_patch.rb

Can provide a fix, but not sure if you're okay with monkey patching Sidekiq 🤷 

(probably worth adding an integration test for this 🤔)

PR #449, opened 2024-01-08
Clarify retry semantics

bdewater:
The way thing were written implied retries were guaranteed - this may be the case when using Sidekiq or when you have configured ApplicationJob with `retry_on StandardError`, but the semantics of both warrant a bit more explanation to understand fully.

Closes https://github.com/Shopify/job-iteration/issues/309
Further expands upon https://github.com/Shopify/job-iteration/pull/433


Mangara (approved):
Nice clarification!

PR #452, opened 2024-01-17
Prepare for isogun removal

Mangara:
The isogun VM is deprecated for local development. This PR makes changes that should work both with and without isogun, to prepare for its eventual demise.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -3,17 +3,16 @@
 name: job-iteration
 
 up:
-  - homebrew:
-    - mysql-client:
-        or:        [mysql@5.7]
-  - ruby:
-      version: 2.7.6
-  - isogun
+  - packages:
+      - mysql_client
+  - ruby: 2.7.6
   - bundler
+  - mysql
+  - redis
   - custom:
       name: Create Job Iteration database
-      meet: mysql -uroot -h job-iteration.railgun -e "CREATE DATABASE job_iteration_test"
-      met?: mysql -uroot -h job-iteration.railgun job_iteration_test -e "SELECT 1" &> /dev/null
+      meet: mysql -uroot -h $MYSQL_HOST -P $MYSQL_PORT -e "CREATE DATABASE job_iteration_test"
+      met?: mysql -uroot -h $MYSQL_HOST -P $MYSQL_PORT job_iteration_test -e "SELECT 1" &> /dev/null
 
 commands:
   test:
--- a/test/support/resque/Rakefile
+++ b/test/support/resque/Rakefile
@@ -11,10 +11,8 @@ require "i18n"
 
 require_relative "../jobs"
 
-redis_host = ENV.fetch("REDIS_HOST") { "localhost" }
-redis_port = ENV.fetch("REDIS_PORT") { 6379 }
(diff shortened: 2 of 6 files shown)


dev.yml
```diff
@@ -3,17 +3,16 @@
...
+      - mysql_client
+  - ruby: 2.7.6
   - bundler
+  - mysql
+  - redis
   - custom:
       name: Create Job Iteration database
-      meet: mysql -uroot -h job-iteration.railgun -e "CREATE DATABASE job_iteration_test"
-      met?: mysql -uroot -h job-iteration.railgun job_iteration_test -e "SELECT 1" &> /dev/null
+      meet: mysql -uroot -h $MYSQL_HOST -P $MYSQL_PORT -e "CREATE DATABASE job_iteration_test"
```
sambostock, line 14: I forget, do we actually need to pass these variables, or does `mysql` just read them automatically?

test/support/resque/Rakefile
```diff
@@ -11,13 +11,10 @@ require "i18n"
 
 require_relative "../jobs"
 
-redis_url = if ENV["USING_DEV"] == "1"
-  "job-iteration.railgun:6379"
-else
-  "localhost:6379"
-end
+redis_host = ENV.fetch("REDIS_HOST") { "localhost" }
+redis_port = ENV.fetch("REDIS_PORT") { 6379 }
```
sambostock, line 15: In what cases would we want the fallback? Also, can't we use `REDIS_URL` directly?

dev.yml
Mangara: We do need to pass them:
```
$ echo $MYSQL_HOST
127.0.0.1
$ echo $MYSQL_PORT
63900
$ mysql -uroot job_iteration_test -e "SELECT 1"
ERROR 2003 (HY000): Can't connect to MySQL server on '127.0.0.1:3306' (61)
$ mysql -uroot -h $MYSQL_HOST -P $MYSQL_PORT job_iteration_test -e "SELECT 1"
+---+
| 1 |
+---+
| 1 |
+---+
```

sambostock:
[The migration guide](https://vault.shopify.io/page/-Isogun~dhb8c8e.md#beta-disabling-isogun:~:text=%E2%9A%A0%EF%B8%8F%20Do%20not%20edit%20or%20remove%20your%20isogun.yml!) explicitly says:

> ⚠️ Do not edit or remove your isogun.yml!

GoodForOneFare:
> 🤷‍♂️ [The migration guide](https://vault.shopify.io/page/-Isogun~dhb8c8e.md#beta-disabling-isogun:~:text=%E2%9A%A0%EF%B8%8F%20Do%20not%20edit%20or%20remove%20your%20isogun.yml!) explicitly says:
>
> ⚠️ Do not edit or remove your isogun.yml!

I think the guidance is there to prevent critical data suddenly disappearing.  e.g., there are likely some `shopify` devs with a curated database stored inside their Isogun container, and they would like a chance to export that.

This seems like a small repo, with a database that's mainly used for tests, so I assume it's safe to 🔥 early.

PR #454, opened 2024-01-23
Explicitly require csv for for Ruby head

Mangara:
Some bundled gems are being removed from Ruby and need to be specified in the Gemfile with Ruby 3.4: https://bugs.ruby-lang.org/issues/20187

This PR adds the CSV gem to our Gemfile for use in testing.

I don't think we need to add it to the Gemspec? If users don't use the CSVEnumerator, they shouldn't need the gem, right?

### DIFF
--- a/lib/job-iteration/csv_enumerator.rb
+++ b/lib/job-iteration/csv_enumerator.rb
@@ -20,7 +20,7 @@ module JobIteration
     #   csv = CSV.open('tmp/files', { converters: :integer, headers: true })
     #   JobIteration::CsvEnumerator.new(csv).rows(cursor: cursor)
     def initialize(csv)
-      unless csv.instance_of?(CSV)
+      unless defined?(CSV) && csv.instance_of?(CSV)
         raise ArgumentError, "CsvEnumerator.new takes CSV object"
       end
 
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ gem "mocha"
 gem "rubocop-shopify", require: false
 gem "yard"
 gem "rake"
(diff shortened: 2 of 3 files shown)


Mangara:
We only use the CSV constant to check that `csv.instance_of?(CSV)`. Should we instead change that to `csv.class.name == "CSV"`?

That would make

1. People trying to call the enumerator without a CSV object get the error

```
CsvEnumerator.new takes CSV object (ArgumentError)
```

2. People (after the first error?) trying to call the enumerator with a CSV object get the error:

```
uninitialized constant CSV (NameError)
```

originating from their own code, which should hopefully get them to `require "csv"` pretty quickly (even Googling it).

The class would still get loaded for apps without the csv gem, but that seems very minor.

lib/job-iteration/csv_enumerator.rb
```diff
@@ -20,7 +20,7 @@ class CsvEnumerator
     #   csv = CSV.open('tmp/files', { converters: :integer, headers: true })
     #   JobIteration::CsvEnumerator.new(csv).rows(cursor: cursor)
     def initialize(csv)
-      unless csv.instance_of?(CSV)
+      unless csv.class.name == "CSV"
```
sambostock, line 23: Another option would be

```suggestion
      unless defined?(CSV) && csv.instance_of?(CSV)
```

or

```suggestion
      unless defined?(CSV) && CSV === csv
```

sambostock (approved):

lib/job-iteration/csv_enumerator.rb
Mangara: Of course! I like that better.

PR #455, opened 2024-02-01
POC - Add custom database role and poc to avoid comparing ids as strings

etravassos:
…rings


lib/job-iteration/active_record_cursor.rb
```diff
@@ -55,26 +55,37 @@ def position=(position)
...
       relation = @base_relation.limit(batch_size)
 
       if (conditions = self.conditions).any?
-        relation = relation.where(*conditions)
+        column = @columns.first.split(".").first
+        model = column.gsub(/_/, " ").split.map(&:capitalize).join("::").singularize.constantize
+
+        attribute = @columns.first.split(".").second.to_sym
+
+        relation = relation.where(model.arel_table[attribute].gt(@position.first))
```
etravassos, line 73: This change is to demonstrate that using `arel_table` does not treat the id column as strings in the generated SQL:

```ruby
Stripe::Fee.where("id > ?", 3).to_sql
=> "SELECT `stripe_fees`.* FROM `stripe_fees` WHERE (id > '3')"

Stripe::Fee.where(Stripe::Fee.arel_table[:id].gt(3)).to_sql
=> "SELECT `stripe_fees`.* FROM `stripe_fees` WHERE `stripe_fees`.`id` > 3"

```

MySql is lenient and allows such comparison, however BigQuery raises an exception for this behaviour:
```
Google::Cloud::InvalidArgumentError: No matching signature for operator > for argument types: INT64, STRING.
```

etravassos: That being said we may want to revisit that coercion on the Mysql queries too 👍 
as you pointed out even though it is lenient about it it ain't ideal.

PR #456, opened 2024-02-02
Use Arel instead of String for AR Enumerator conditionals

pedropb:
# Problem statement

On our application we noticed some inconsistencies with the usage of Rails 7.1 and job-iteration.

For example:
```ruby
enumerator_builder.active_record_on_batches(Model.all, cursor: cursor)
```

when iterating would result in queries like:
```sql
 WHERE (reconciled_transaction_fees.id > '238427110') ORDER BY reconciled_transaction_fees.id LIMIT 3
 WHERE (reconciled_transaction_fees.id > '304256793') ORDER BY reconciled_transaction_fees.id LIMIT 3
 WHERE (reconciled_transaction_fees.id > '421883103') ORDER BY reconciled_transaction_fees.id LIMIT 3
```

While `reconciled_transaction_fees.id > '421883103'` is not wrong for adapters like MySQL and Postgres, it relies on behaviour from those DBMS to convert from strings to integer and still use the indices.

While working on a BigQuery adapter for ActiveRecord, we noticed that BigQuery will raise when comparing numeric columns with strings.

# Proposal

So instead of manipulating strings to build the paging conditional:
https://github.com/Shopify/job-iteration/blob/main/lib/job-iteration/active_record_batch_enumerator.rb#L97-L101

We propose using Rails Arel:
https://github.com/Shopify/job-iteration/blob/5d18040d0b12350089b03b4e57f908b84dbb5b57/lib/job-iteration/active_record_cursor.rb#L82-L86

which will emit queries without the quotes:
```sql
WHERE `reconciled_transaction_fees`.`id` > 238427110 ORDER BY `reconciled_transaction_fees`.`id` LIMIT 3
WHERE `reconciled_transaction_fees`.`id` > 304256793 ORDER BY `reconciled_transaction_fees`.`id` LIMIT 3
WHERE `reconciled_transaction_fees`.`id` > 421883103 ORDER BY `reconciled_transaction_fees`.`id` LIMIT 3
```

# What's in this PR?

These changes affect:
- enumerator_builder.active_record_on_records
- enumerator_builder.active_record_on_batches

The change affects only the `columns:` optional argument.
- `columns:` can still be passed as strings or symbols. Internally, the job-iteration code will convert the symbols to Arel attributes and manipulate them as Arel instead of strings.
~- (new!) `columns:` can also be passed as Arel attributes.~

# The Ugly

There are 2 main problems with the approach:
1) Arel is a private API subject to change. I tested this with Rails 7 (the job-iteration unit tests) and 7.1 (my app's unit tests). I'm relying on CI to test against other builds.

2) The new code **does not support** multi-column cursor when more than one table is used (`cursor: ["products.id", "comments.id"]`). While trying to add support to that, I created a test file and noticed that  job-iteration code had a bug for multi-column cursors on different tables. This PR does not resolve the bug (#457 )

### DIFF
--- a/lib/job-iteration/active_record_cursor.rb
+++ b/lib/job-iteration/active_record_cursor.rb
@@ -18,12 +18,8 @@ module JobIteration
       end
     end
 
-    def initialize(relation, columns = nil, position = nil)
-      @columns = if columns
-        Array(columns)
-      else
-        Array(relation.primary_key).map { |pk| "#{relation.table_name}.#{pk}" }
-      end
+    def initialize(relation, columns, position = nil)
+      @columns = columns
       self.position = Array.wrap(position)
       raise ArgumentError, "Must specify at least one column" if columns.empty?
       if relation.joins_values.present? && !@columns.all? { |column| column.to_s.include?(".") }
@@ -34,7 +30,7 @@ module JobIteration
         raise ConditionNotSupportedError
       end
 
-      @base_relation = relation.reorder(@columns.join(","))
+      @base_relation = relation.reorder(*@columns)
       @reached_end = false
     end
 
@@ -54,12 +50,10 @@ module JobIteration
 
     def update_from_record(record)
       self.position = @columns.map do |column|
-        method = column.to_s.split(".").last
-
-        if ActiveRecord.version >= Gem::Version.new("7.1.0.alpha") && method == "id"
+        if ActiveRecord.version >= Gem::Version.new("7.1.0.alpha") && column.name == "id"
           record.id_value
         else
-          record.send(method.to_sym)
+          record.send(column.name)
         end
       end
     end
@@ -89,14 +83,14 @@ module JobIteration
       i = @position.size - 1
       column = @columns[i]
       conditions = if @columns.size == @position.size
-        "#{column} > ?"
+        column.gt(@position[i])
       else
-        "#{column} >= ?"
+        column.gteq(@position[i])
       end
       while i > 0
         i -= 1
         column = @columns[i]
-        conditions = "#{column} > ? OR (#{column} = ? AND (#{conditions}))"
+        conditions = column.gt(@position[i]).or(column.eq(@position[i]).and(conditions))
       end
       ret = @position.reduce([conditions]) { |params, value| params << value << value }
       ret.pop
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
 when generating position for cursor based on `:id` column (Rails 7.1 and above, where composite
 primary models are now supported). This ensures we grab the value of the id column, rather than a
 potentially composite primary key value.
+- [456](https://github.com/Shopify/job-iteration/pull/431) - Use Arel to generate SQL that's type compatible for the
+  cursor pagination conditionals in ActiveRecord cursor. Previously, the cursor would coerce numeric ids to a string value 
+  (e.g.: `... AND id > '1'`)
 
 ## v1.4.1 (Sep 5, 2023)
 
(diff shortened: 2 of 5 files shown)


lib/job-iteration/active_record_cursor.rb
```diff
@@ -18,23 +18,16 @@ def initialize
...
-        Array(columns)
-      else
-        Array(relation.primary_key).map { |pk| "#{relation.table_name}.#{pk}" }
-      end
+    def initialize(relation, columns, position = nil)
+      @columns = columns
       self.position = Array.wrap(position)
       raise ArgumentError, "Must specify at least one column" if columns.empty?
-      if relation.joins_values.present? && !@columns.all? { |column| column.to_s.include?(".") }
-        raise ArgumentError, "You need to specify fully-qualified columns if you join a table"
```
nvasilevski, line 30: Does it mean that for a multi-table columns the only option is to pass arel attribute instances? I think we may want to keep the `"<table_name>.<column_name>"` as the only accepted format to avoid forcing applications to depend on Arel

nvasilevski (commented):
>Arel is a private API subject to change.

Fair but it feels like building raw SQL is even less stable especially when it comes to supporting multiple db adapters

>(new!) columns: can also be passed as Arel attributes.

But I wonder if that's something we should still disallow. One thing is to rely on Arel internally but the other thing is to allow applications to pass arel attributes and promise that the gem will always (within a major release) maintain it. Perhaps it should actually be disallowed 

>The new code does not support  multi-column cursor when more than one table is used

But it wasn't supported before, right? Or at least it wasn't working due to a bug, right? So I'm not too concerned about it as long as it doesn't break something that used to be working

Overall I think this is a reasonable change. @Shopify/rails would love to hear more opinions!

Mangara (commented):
I agree with Nikita - using Arel internally in the cursor is fine (assuming we have good test coverage for it), but the interface for users of the gem should just accept strings and symbols.

lib/job-iteration/active_record_cursor.rb
pedropb: I reverted this change. The guard clause is back in place now.

pedropb:
Thanks for the feedback @nvasilevski @Mangara 
I think it's a fair compromise to keep the Arel dependency internal to job-iteration and not expose/support it publicly.

Pushed commit 51d8f8e0433357a8d77187ee167f2366548277be to drop the conditional that handled Arel inputs as suggested.

Mangara:
If it's important to prevent regressions, I'm fine with testing the precise SQL queries we generate. Ideally, we'd test the public interface of `ActiveRecordEnumerator` instead of `ActiveRecordCursor` or the `conditions` method directly.

nvasilevski (approved):

Mangara:
Oh, before we merge - could you add a note to the change log?

pedropb:
Change log added!
Merging.

lib/job-iteration/active_record_cursor.rb
```diff
@@ -18,12 +18,8 @@ def initialize
...
-      @columns = if columns
-        Array(columns)
-      else
-        Array(relation.primary_key).map { |pk| "#{relation.table_name}.#{pk}" }
-      end
+    def initialize(relation, columns, position = nil)
+      @columns = columns
       self.position = Array.wrap(position)
       raise ArgumentError, "Must specify at least one column" if columns.empty?
       if relation.joins_values.present? && !@columns.all? { |column| column.to_s.include?(".") }
```
segiddins, line 25: is this still necessary, now that it's an arel attribute being passed in instead of a string?

pedropb: This logic was left untouched because on the presence of join conditions, the ORDER by columns can become ambiguous and fails the queries.
The `.all?` clause was a constraint to prevent this failure mode.

Also FWIW, the changes in this PR were reverted in #483 (v1.5.1). There's more context about the bug it introduced in #483.

PR #459, opened 2024-02-02
Add Rails integration test

sambostock:
The purpose of this test is to exercise the happy path of running a job in the context of a full Rails app, including the upcoming Railtie we'll be adding in #441.

### DIFF
--- a/test/integration/rails_integration_test.rb
+++ b/test/integration/rails_integration_test.rb
@@ -1,0 +1,122 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "tempfile"
+require "fileutils"
+require "open3"
+
+class RailsIntegrationTest < ActiveSupport::TestCase
+  CommandFailed = Class.new(StandardError)
+
+  test "running a job in a Rails app works" do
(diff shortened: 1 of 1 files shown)


test/integration/rails_integration_test.rb
```diff
@@ -0,0 +1,122 @@
...
+            run_or_raise("gem", "install", "nio4r")
+            run_or_raise("gem", "install", "websocket-driver")
+            run_or_raise("gem", "install", "date")
+            run_or_raise("gem", "install", "racc")
+
+            # It is also unclear why we need to --skip-install, lock, and install, instead of just using `bundle add`,
+            # but, again, otherwise it blows up with the same error as above.
+            run_or_raise("bundle", "add", "rails", "--version", rails_version.to_s, "--skip-install")
+            run_or_raise("bundle", "lock")
+            run_or_raise("bundle", "install")
```
sambostock, line 56: There's something really weird going on here. I had to add all of this for it to run locally, but we still run into this error in CI, just for different gems.

I think there's something weird going on, to do with Bundler/Rubygems environment variables or something, and the interaction with running `bundler` within this app.

PR #460, opened 2024-02-09
Add tests for `JobIteration::TestHelper`

sambostock:
This adds tests for [`JobIteration::TestHelper`](https://github.com/Shopify/job-iteration/blob/main/lib/job-iteration/test_helper.rb), which was previously untested.

### DIFF
--- a/test/unit/test_helper_test.rb
+++ b/test/unit/test_helper_test.rb
@@ -1,0 +1,106 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+module JobIteration
+  class TestHelperTest < ActiveSupport::TestCase
+    include JobIteration::TestHelper
+
+    class CounterJob < ActiveJob::Base
+      include JobIteration::Iteration
+
+      class << self
+        attr_writer :count
+
+        def count
+          @count ||= 0
+        end
+      end
+
+      def build_enumerator(times, cursor:)
+        enumerator_builder.array(Array.new(times), cursor: cursor)
+      end
+
+      def each_iteration(_, _)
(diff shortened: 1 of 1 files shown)


sambostock:
Something really weird is going on. This PR only adds tests, but that seems to introduce flakiness into the rest of the test suite which consistently causes failures in at least one of the test runs across the matrix. There does not appear to be a pattern in terms of _which_ matrix entries fail.

The only thing I can think of is that either
- something in the test file's code is leaking, or
- something in `JobIteration::TestHelper` is leaking.

test/unit/test_helper_test.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    test "#continue_iterating allows jobs to iterate until the end" do
+      iterate_exact_times(3.times)
+      continue_iterating
+
+      CounterJob.perform_now(10)
+
+      assert_equal 10, CounterJob.count
+    end
+
```
Mangara, line 58: There are two extra cases I'd like to add, based on the test that failed in core:

```suggestion
    test "#iterate_once allows jobs to run one iteration at a time" do
      iterate_once

      job = CounterJob.perform_now(10)
      job.perform_now

      assert_equal 2, CounterJob.count
    end

    test "#continue_iterating allows the job to finish after running initial iterations" do
      iterate_once

      job = CounterJob.perform_now(10)

      continue_iterating

      job.perfom_now

      assert_equal 10, CounterJob.count
    end
```

Mangara (approved):
Thanks for writing these!

PR #472, opened 2024-04-02
Allow instance of ActiveRecord::Base in array enumerators

acaron:
I'm removing the check on ActiveRecord::Base instance in the array enumerator to allow single AR records. There is a check already on Array so no ActiveRelation::Relation are given as the base enumerator. From a discussion with @kirs, this check is too defensive in the case that someone accidently tries to use the array enumerator on AR relation.


kirs (approved):
I was the one who've added that and now a few years since, I agree that's too defensive.
It was there to protect developers from shooting into the foot and using array enumerator when they actually needed AR enumerator.

But it's called `build_array_enumerator` anyways, so there's the intention of working with an array.

PR #504, opened 2024-09-23
Add change log entry for GoodJob support

Mangara:
Adds a change log entry for https://github.com/Shopify/job-iteration/pull/464 and fixes CI.

CC @olivier-thatch


olivier-thatch:
Thanks @Mangara!

Issue #496, opened 2024-08-19
Support SolidQueue


tmimura39:
Hi, I am currently using this Gem with Sidekiq as backend. (To be precise, we use maintenance_tasks.)

We are considering moving from Sidekiq to [SolidQueue](https://github.com/rails/solid_queue)
[SolidQueue will be the default backend for ActiveJob in Rails v8](https://github.com/rails/rails/issues/50442)

It appears that the job-iteration gem does not currently define an  interruption_adapter for SolidQueue.
How about adding SolidQueue's interruption_adapter in the job-iteration gem to help many people?

This is the kind of content we're expecting.
```ruby
# frozen_string_literal: true

begin
  require "solid_queue"
rescue LoadError
  # SolidQueue is not available, no need to load the adapter
  return
end

begin
  # SolidQueue.on_worker_stop was introduced in SolidQueue 0.7.1
  gem("solid_queue", ">= 0.7.1")
rescue Gem::LoadError
  warn("job-iteration's interruption adapter for SolidQueue requires SolidQueue 0.7.1 or newer")
  return
end

module JobIteration
  module InterruptionAdapters
    module SolidQueueAdapter
      class << self
        attr_accessor :stopping

        def call
          stopping
        end
      end
    end

    SolidQueue.on_worker_stop do
      JobIteration::Integrations::SolidQueueAdapter.stopping = true
    end

    register(:solid_queue, SolidQueueAdapter)
  end
end
```

tmimura39:
We are having SolidQueue add the functionality to the SolidQueue side.
Soon you will be able to create an optimal SolidQueueAdapter.

https://github.com/rails/solid_queue/issues/282#issuecomment-2327657330

PR #505, opened 2024-09-23
Add interruption adapter for Solid Queue

Mangara:
Resolves https://github.com/Shopify/job-iteration/issues/496

This PR adds an interruption adapter for [Solid Queue](https://github.com/rails/solid_queue). This will be the default Active Job back-end starting in Rails 8, so it's important that job-iteration works with it out of the box.

We can't test the loading properly, because even requiring `solid_queue` fails if we don't have `::Rails::Engine` and `ActiveModel` loaded.

TODO:
- [x] Rebase once https://github.com/Shopify/job-iteration/pull/504 merges
- [x] Manual test in a Rails app with Solid Queue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### Main (unreleased)
 
 - [464](https://github.com/Shopify/job-iteration/pull/464) - Add interruption adapter for [GoodJob](https://github.com/bensheldon/good_job).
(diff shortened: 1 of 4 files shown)


CelinaAssal (approved):

exterm:
Should this be added to the docs at https://github.com/Shopify/job-iteration?tab=readme-ov-file#requirements ?

(👋🏼 @Mangara )

PR #509, opened 2024-09-29
Enable CSV Batching

harmanT23:
I noticed that although CSV batching was implemented in [CsvEnumerator](https://github.com/Shopify/job-iteration/blob/d561d7e1a88c456689b1918f065609d264ec3c09/lib/job-iteration/csv_enumerator.rb#L41-L47), it was not implemented yet in `EnumeratorBuilder`.

I've gone ahead and added in `build_csv_enumerator_on_batches` with a default `batch_size` of `100` identical to the default used for active record batching along with an alias of `csv_on_batches`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ### Main (unreleased)
 
-Nil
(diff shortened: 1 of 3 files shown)


Mangara (approved):
Nice! Can you add this to the change log under unreleased?

A rebase should fix the CI failure.

PR #512, opened 2024-10-09
Fix the UTC timezone issue with cursor resume

yigitozkavci:


CelinaAssal (approved):

Mangara (approved):
Just missing a change to `enumerator_builder` to pass through the `timezone`, LGTM otherwise.

Mangara:
> Just missing a change to enumerator_builder to pass through the timezone

Never mind, the splats take care of this. Good to merge!

lib/job-iteration/active_record_enumerator.rb
```diff
@@ -61,6 +62,7 @@ def column_value(record, attribute)
       value = record.read_attribute(attribute.to_sym)
       case record.class.columns_hash.fetch(attribute).type
       when :datetime
+        value = value.in_time_zone(@timezone) unless @timezone.nil?
```
bdewater, line 65: Not sure if this has to be configured explicitly. Active Job [serializes](https://github.com/rails/rails/blob/fe0fb9fdf04f638996e785d3e291f5fe8bc138a0/activejob/lib/active_job/core.rb#L118) and [deserializes](https://github.com/rails/rails/blob/fe0fb9fdf04f638996e785d3e291f5fe8bc138a0/activejob/lib/active_job/core.rb#L159) the timezone on the job, and [sets it](https://github.com/rails/rails/blob/fe0fb9fdf04f638996e785d3e291f5fe8bc138a0/activejob/lib/active_job/timezones.rb#L9) in an `around_perform` hook. Could this simply be:
```rb
value.in_time_zone(::Time.zone)
```

Mangara: Isn't that the default behaviour? I think the original issue is with a job enqueued (and running) in the local timezone, but reading data that was stored in a different time zone (UTC).

PR #515, opened 2024-10-25
Correctly calculate size for array enumerator

bdewater:
I don't think this is an actual problem for anybody, but if we're passing a size to `to_enum` it should be correct. Example:
```rb
array = ["a", "b", "c"]
drop = 1

current = array.each_with_index.drop(drop).to_enum { enumerable.size } 
current.size # 3 - wrong!
current.to_a.size # 2

new = array.each_with_index.drop(drop).to_enum { enumerable.size - drop } 
new.size # 2
```

### DIFF
--- a/lib/job-iteration/enumerator_builder.rb
+++ b/lib/job-iteration/enumerator_builder.rb
@@ -62,7 +62,7 @@ module JobIteration
           cursor + 1
         end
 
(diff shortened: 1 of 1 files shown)


Mangara (approved):
Good catch!

PR #533, opened 2025-01-15
Add Tapioca compiler

st0012:


lib/tapioca/dsl/compilers/job_iteration.rb
```diff
@@ -0,0 +1,96 @@
...
+              "perform_later",
+              parameters: expanded_parameters,
+              return_type: "T.any(#{constant_name}, FalseClass)",
+              class_method: true,
+            )
+
+            job.create_method(
+              "perform_now",
+              parameters: expanded_parameters,
+              return_type: return_type,
```
Mangara, line 51: The return type of `build_enumerator` will generally be an enumerator of some kind. The return type of` perform_now` is `void` for job-iteration jobs: https://github.com/Shopify/job-iteration/blob/7c808419d82860baf3aefaff9b32e08e5c415afb/lib/job-iteration/iteration.rb#L113

lib/tapioca/dsl/compilers/job_iteration.rb
```diff
@@ -0,0 +1,96 @@
+# typed: strict
+# frozen_string_literal: true
+
+return unless defined?(JobIteration::Iteration)
+
+require "prism"
```
paracycle, line 6: We should not need Prism in DSL compilers at all. Why do we need this?

lib/tapioca/dsl/compilers/job_iteration.rb
```diff
@@ -0,0 +1,96 @@
...
+              else
+                typed_param
+              end
+            end
+
+            return_type = compile_method_return_type_to_rbi(method)
+
+            job.create_method(
+              "perform_later",
+              parameters: expanded_parameters,
```
Mangara, line 43: We're missing the potential block here: https://github.com/Shopify/tapioca/blob/91ad27251ad8c422f116cd0eafe9a5bd82ea7d70/lib/tapioca/dsl/compilers/active_job.rb#L32

lib/tapioca/dsl/compilers/job_iteration.rb
st0012: Updated.

lib/tapioca/dsl/compilers/job_iteration.rb
```diff
@@ -0,0 +1,120 @@
...
+            end
+          end
+        end
+
+        class << self
+          extend T::Sig
+
+          sig { override.returns(T::Enumerable[Module]) }
+          def gather_constants
+            all_classes.select { |c| c < ::JobIteration::Iteration }
```
paracycle, line 114:
```suggestion
            descendants_of(::JobIteration::Iteration)
```

lib/tapioca/dsl/compilers/job_iteration.rb
```diff
@@ -0,0 +1,120 @@
...
+            parameters + [create_block_param(
+              "block",
+              type: "T.nilable(T.proc.params(job: #{constant_name}).void)",
+            )]
+          else
+            parameters
+          end
+        end
+
+        def compile_method_parameters_to_rbi(method_def)
```
paracycle, line 68: Do you need to copy this complex method over? I would have instead done a call to `compile_method_parameters_to_rbi` and then walked over the result to find all required params of type `T::Types:FixedHash` and replace them with the keyword parameters instead.

lib/tapioca/dsl/compilers/job_iteration.rb
paracycle: Ah, right, sorry. In that case:
```suggestion
            all_classes.select { |c| ::JobIteration::Iteration > c }
```
otherwise, there might be a world of pain waiting for you.

(Looking at you `XPath`: https://github.com/Shopify/tapioca/pull/1969)

lib/tapioca/dsl/compilers/job_iteration.rb
paracycle: Ah, sorry, those are also strings. I guess you could do:
```ruby
signature = signature_of(method_def)
fixed_hash_args = signature.arg_types.select { |arg_type| arg_type[1].is_?(T::Types::FixedHash) }
```
and loop over those to match the params to the ones you need to replace.

Mangara (approved):
LGTM once the CI failures are addressed.

paracycle (approved):

PR #534, opened 2025-01-16
Fix CI

st0012:
`ActiveSupport::LoggerThreadSafeLevel` depends on `logger`, but doesn't require it explicitly prior to 7.1. In the real application, the `logger` is likely required by other components so this is not an issue. But for this gem's test setup, we need to require it explicitly.


Mangara (approved):
Thank you for fixing this!

PR #536, opened 2025-01-22
Support keyword arguments in iteration jobs

st0012:
It's now more common to have jobs that take keyword arguments, like `MyJob.perform_later(foo: "bar")`. But under the current implementation, `build_enumerator` can only take them as a hash argument instead of keyword arguments:

```ruby
  foo = params[:foo]
  # ...
end
```

This commit improves that by allowing keyword arguments to be passed to `build_enumerator` and `each_iteration` as:

```ruby
def build_enumerator(foo:, cursor:)
  # ...
end
```

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -107,8 +107,8 @@ module JobIteration
       self.total_time = Float(job_data["total_time"] || 0.0)
     end
 
-    def perform(*params) # @private
-      interruptible_perform(*params)
+    def perform(*args, **kwargs) # @private
+      interruptible_perform(*args, **kwargs)
 
       nil
     end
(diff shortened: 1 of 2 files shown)


Mangara:
@st0012 Can you explain why you closed this? Does it not work? Or was it just unnecessary? Just looking to get some context if someone asks about this in the future.

st0012:
I closed this because I later realised that the Shopify apps we care about actually can't adopt the keyword arguments pattern directly. And while this support won't negatively impact those apps either, having it would require `job-iteration` dropping Ruby 2.6 & 2.7 support at the same time.
So instead of dropping Ruby support for a feature nobody is asking for, I decided to close it and let others to pick it up if needed.

PR #537, opened 2025-01-23
Correct return type for `perform_now`

st0012:
If the job has registered an exception handler, like `discard_on MyError`, when `perform_now` is called and triggered `MyError`, an instance of `MyError` is returned.

Source: https://github.com/rails/rails/blob/b1ec311a8fa3b8e1ac5cdcfd850b793f8cf5aecd/activejob/lib/active_job/execution.rb#L54

This is an intentional behaviour from ActiveJob, so we should respect it and update the return type of `perform_now` to be `T.any(NilClass, Exception)`.


Mangara (approved):
Good find! I had no idea this worked this way.

PR #538, opened 2025-01-28
[Tapioca compiler] Add nil default value to nilable parameters

st0012:
When a parameter is nilable, the generated keyword parameter should have a `nil` default value. This reduces the amount of changes needed when adopting the generated type signatures.

For example, with this job:

```rb
class NotifyJob < ActiveJob::Base
  Params = T.type_alias { { user_id: T.nilable(Integer), name: String } }
  extend T::Sig
  sig { params(params: Params, cursor: T.untyped).returns(T::Array[T.untyped]) }
    # ...
  end
end
```

currently the generated signature would be:

```rbi
sig { params(user_id: T.nilable(::Integer), name: ::String, block: T.nilable(T.proc.params(job: NotifyJob).void)).returns(T.any(NotifyJob, FalseClass)) }
def perform_later(user_id:, name:, &block); end
```

which would require all callers to pass the `user_id` param even when it was not previously needed.

But with this change, the generated signature would be:

```rb
sig { params(name: ::String, user_id: T.nilable(::Integer), block: T.nilable(T.proc.params(job: NotifyJob).void)).returns(T.any(NotifyJob, FalseClass)) }
def perform_later(name:, user_id: nil, &block); end
```

which will allow users to skip the `user_id` param unless when it's needed.

### DIFF
--- a/lib/tapioca/dsl/compilers/job_iteration.rb
+++ b/lib/tapioca/dsl/compilers/job_iteration.rb
@@ -10,6 +10,15 @@ module Tapioca
         extend T::Sig
 
         ConstantType = type_member { { fixed: T.class_of(::JobIteration::Iteration) } }
+        PARAM_TYPES_IN_ORDER = [
+          RBI::Param,
+          RBI::OptParam,
+          RBI::RestParam,
+          RBI::KwParam,
+          RBI::KwOptParam,
+          RBI::KwRestParam,
+          RBI::BlockParam,
+        ].freeze
 
         sig { override.void }
         def decorate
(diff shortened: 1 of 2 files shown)


lib/tapioca/dsl/compilers/job_iteration.rb
```diff
@@ -39,6 +52,9 @@ def decorate
               expanded_parameters = parameters
             end
 
+            # Sorbet expects optional keyword arguments to be after required keyword arguments.
+            expanded_parameters.sort_by! { |typed_param| PARAM_TYPES_IN_ORDER.index(typed_param.param.class) }
+
```
paracycle, line 57: This is fine for now, but I wonder if `rbi` gem should be responsible for generating a method signature to match this ordering, in the future.

@Morriar WDYT?

paracycle (approved):

Mangara (approved):
Nice work!

Issue #550, opened 2025-03-07
Custom Tapioca compiler does not consider generic jobs


emilylin0:
Hi!

We've been trying to bump the job-iteration gem from 1.5.1 to 1.9.0, but we've run into some [Sorbet failures](https://buildkite.com/shopify/russbot/builds/59465#01957281-70e7-4024-aa44-e164a2a0380d) due to the custom Tapioca compiler that was added in [this PR ](https://github.com/Shopify/job-iteration/pull/533).

We have a number of jobs that extend `T::Generic`, but because of this, we've been getting errors like the one below.

Would it be possible to update the compiler so that it generates the correct signature if there are generic jobs? 

Thanks in advance! 🙏 

### Sorbet error for `perform_later` signature

````
sorbet/rbi/dsl/controls/netsuite_documents_have_been_pushed.rbi:23: Malformed type declaration. Generic class without type arguments Controls::NetsuiteDocumentsHaveBeenPushed https://srb.help/5045
  | 23 \|      ).returns(T.any(Controls::NetsuiteDocumentsHaveBeenPushed, FalseClass))
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | Autocorrect: Use -a to autocorrect
  | sorbet/rbi/dsl/controls/netsuite_documents_have_been_pushed.rbi:23: Replace with 
Controls::NetsuiteDocumentsHaveBeenPushed[T.untyped]
  | 23 \|      ).returns(T.any(Controls::NetsuiteDocumentsHaveBeenPushed, FalseClass))
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |  
  | sorbet/rbi/dsl/controls/netsuite_documents_have_been_pushed.rbi:22: Malformed type declaration. Generic class without type arguments Controls::NetsuiteDocumentsHaveBeenPushed https://srb.help/5045
  | 22 \|        block: T.nilable(T.proc.params(job: Controls::NetsuiteDocumentsHaveBeenPushed).void)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | Autocorrect: Use -a to autocorrect
  | sorbet/rbi/dsl/controls/netsuite_documents_have_been_pushed.rbi:22: Replace with Controls::NetsuiteDocumentsHaveBeenPushed[T.untyped]
[...]

PR #556, opened 2025-03-18
Update sorbet compiler so that it correctly handles generic jobs

emilylin0:
Issue: https://github.com/Shopify/job-iteration/issues/550#issuecomment-2707613910

tl;dr is that some repos define generic jobs, but the compiler wasn't creating the correct signature


Mangara (approved):
Can you test this branch in your app, to make sure the generated types are correct?

emilylin0:
Looks like Sorbet isn't complaining anymore

Screenshot 2025-03-18 at 2 26 47 PM

Once this gets shipped, is it possible to cut a new release with these changes?

st0012 (approved):

ehasrouni (approved):

PR #594, opened 2025-07-14
Clean up CSVEnumerator

Mangara:
This PR removes a work-around for Ruby 2.6 that is no longer necessary. It also ensures that calling `size` on a batch CSV enumerator returns `nil` if the file path cannot be found, instead of `0`.


MatthewRBruce (approved):

Mangara:
The Rails 7.1 CI failures are caused by https://github.com/sidekiq/sidekiq/issues/6746 which will be fixed in the next Sidekiq release.

PR #606, opened 2025-07-30
fix: Make `EnumeratorBuilder#wrap` match documentation

ixti:
As of #513 enums must be wrapped, and documentation shows examples like:

``` ruby
def build_enumerator(cursor:)
  enum_builder.wrap(
    some_enumerator
  )
end
```

However, `EnumeratorBuilder#wrap` is delegated to `@wrapper` which expects builder's instance as first argument. Thus, without this PR's patch we actually have to write code like:

``` ruby
def build_enumerator(cursor:)
  enum_builder.wrap(
    enum_builder,
    some_enumerator
  )
end
```

### DIFF
--- a/lib/job-iteration/enumerator_builder.rb
+++ b/lib/job-iteration/enumerator_builder.rb
@@ -35,18 +35,20 @@ module JobIteration
       @wrapper = wrapper
     end
 
-    def_delegator :@wrapper, :wrap
+    def wrap(enum)
+      @wrapper.wrap(self, enum)
+    end
 
     # Builds Enumerator objects that iterates once.
     def build_once_enumerator(cursor:)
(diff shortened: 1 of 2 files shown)


kindjar:
I am currently trying to resolve a deprecation warning about an unwrapped enumerator and discovered the same documentation/API mismatch. If I change my code to match the documented API, it won't work with the current version of the gem, but if I were to fix it to match the _actual_ current implementation, then it doesn't match the docs and if this PR (which I personally, think is better than the current _actually implemented_ API for `wrap`, for what it's worth) gets merged, it would break my "fixed" code. It would be really nice to get some feedback from the maintainers about what the intent is for the fix here, if y'all have a moment to chime in! ❤️

PR #690, opened 2026-03-25
WIP - parallel array iteration

Mangara:
### What and why?

There are cases where there is so much work to iterate over that it makes sense to do it in parallel. It's possible to do this outside of job-iteration, but this prototype explores what a solution inside job-iteration could look like.


Mangara:
Closing in favour of https://github.com/Shopify/job-iteration/pull/702

PR #699, opened 2026-04-23
Fix Tapioca Compiler to Preserve Params for Empty FixedHash

beepubapu:
## Summary

  The tapioca DSL compiler for JobIteration explodes FixedHash types into keyword arguments on the generated `perform`/`perform_now`/`perform_later` methods. When the FixedHash is empty (`{}`),
  `hash_type.types.map` returns `[]`, and `flat_map` replaces the positional `params` parameter with nothing — generating zero-arg methods.

  This means a job like:

  ```ruby
  sig { params(params: {}, cursor: T.untyped).returns(T::Array[T.untyped]) }

  # generates:

  # Before (broken)
  def perform; end
  def perform_now; end
  ```

  Any call site using perform_now({}) then fails with Sorbet error 7004 (too many arguments).

  ## Fix

When the FixedHash has zero keys, preserve the original positional parameter as-is instead of exploding to nothing:
  
```ruby                                                                                                                                                                                                       
  # After                                                                                                                                                                                                      
  sig { params(params: {}).void }
  def perform(params); end                                                                                                                                                                                       
  ```
                                                                                                                                                                                                                 
  The param stays required because the runtime method chain (`perform(*params) → build_enumerator(*arguments, cursor:)`) splats caller arguments directly — an RBI default does not exist at runtime, so `perform_now()` with zero args would raise `ArgumentError` in `build_enumerator`. Callers must pass `perform_now({})`.

Non-empty FixedHash explosion is unchanged — all existing behavior is preserved.

Also extracted `expand_fixed_hash` private method to keep block nesting within rubocop's 3-level limit (`Metrics/BlockNesting`).

## Tests

  - Added `test_generates_correct_rbi_file_for_job_with_empty_fixed_hash_parameter` — verifies the fix
  - Added `test_generates_correct_rbi_file_for_job_with_all_nilable_fixed_hash_keys` — verifies all-optional keys still explode to optional kwargs (no regression)

### DIFF
--- a/lib/tapioca/dsl/compilers/job_iteration.rb
+++ b/lib/tapioca/dsl/compilers/job_iteration.rb
@@ -86,7 +86,7 @@ module Tapioca
         sig { params(typed_param: RBI::TypedParam, hash_type: T::Types::FixedHash).returns(T.any(RBI::TypedParam, T::Array[RBI::TypedParam])) }
         def expand_fixed_hash(typed_param, hash_type)
           if hash_type.types.empty?
-            create_opt_param(typed_param.param.name, type: "{}", default: "{}")
+            typed_param
           else
             hash_type.types.map do |key, value|
               if value.name.start_with?("T.nilable")
--- a/lib/tapioca/dsl/compilers/job_iteration.rb
+++ b/lib/tapioca/dsl/compilers/job_iteration.rb
@@ -38,13 +38,7 @@ module Tapioca
               fixed_hash_args = signature.arg_types.select { |arg_type| T::Types::FixedHash === arg_type[1] }.to_h
               expanded_parameters = parameters.flat_map do |typed_param|
                 if (hash_type = fixed_hash_args[typed_param.param.name.to_sym])
-                  hash_type.types.map do |key, value|
-                    if value.name.start_with?("T.nilable")
-                      create_kw_opt_param(key.to_s, type: value.to_s, default: "nil")
-                    else
-                      create_kw_param(key.to_s, type: value.to_s)
-                    end
-                  end
+                  expand_fixed_hash(typed_param, hash_type)
                 else
                   typed_param
                 end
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by liamstewart — APPROVED** [review]

**Comment by beepubapu** [comment]
## lib/tapioca/dsl/compilers/job_iteration.rb

### Line Comment (line 89):
🐛 **Bug**: The new empty-`FixedHash` branch does not preserve the original parameter shape and always replaces the original parameter with `create_opt_param(..., default: "{}")`. For the case described in the PR, a method like `def build_enumerator(params, cursor:)` now generates RBI methods such as `def perform(params = {}); end`, `def perform_now(params = {}); end`, and `def perform_later(params = {}); end`. This makes Sorbet accept `MyJob.perform_now()` and `MyJob.perform_later()` with no arguments. However, the runtime still forwards only the provided arguments into `build_enumerator` via `perform(*params)` -> `interruptible_perform(*params)` -> `build_enumerator(*arguments, cursor: cursor_position)`. A zero-argument call therefore becomes `build_enumerator(cursor: ...)`, which raises because `params` remains a required positional argument. The runtime tests in the repository also reflect this contract by invoking such jobs as `perform_now({})`, not `perform_now()`. This branch also discards the original parameter kind and order: if the original empty fixed-hash parameter was required or keyword-based, it is still rewritten into an optional positional param, which can further drift from the real `build_enumerator` API. The result is an RBI that advertises call forms the runtime cannot safely handle, potentially causing runtime exceptions, failed async jobs, and retry churn.

**Suggestion:** Preserve the original `typed_param` for empty fixed hashes instead of always synthesizing a new optional positional param:

```ruby
if hash_type.types.empty?
  typed_param
else
  hash_type.types.map do |key, value|
    if value.name.start_with?("T.nilable")
      create_kw_opt_param(key.to_s, type: value.to_s, default: "nil")
    else
      create_kw_param(key.to_s, type: value.to_s)
    end
  end
end
```

If special handling is needed for genuinely optional empty-hash params, branch on `typed_param.param.class` rather than unconditionally converting every empty fixed-hash param into `RBI::OptParam`.

SomalianIvan (approved):

Mangara (approved):

PR #702, opened 2026-04-28
Add parallel iteration scaffold

Mangara:
### What and why?

It's sometimes useful to iterate over a collection in parallel to speed up the completion of the work. This PR adds a primitive to enable that:

```ruby
def build_enumerator(cursor:)
  enumerator_builder.parallel(instances: 7, cursor: cursor) do |instance, total_instances, inner_cursor|
    enumerator_builder.active_record_on_records(
      Product.where("id % ? = ?", total_instances, instance),
      cursor: inner_cursor,
    )
  end
end
```

When this job is enqueued, it first runs with a `nil` cursor, causing it to enqueue the 5 child jobs, each with a cursor that looks like `{ instance: x, inner_cursor: nil }`, where `x` ranges from 0 to 4. When those jobs start, they build the inner enumerator and run as normal, except that the outer enumerator ensures that the cursor stays wrapped in this hash with `instance` and `inner_cursor`.

```mermaid
sequenceDiagram
      participant Parent as ParentJob
      participant Q as ActiveJob queue
      participant Child as ChildJob (instance i)

      activate Parent
      Parent->>Parent: build_enumerator(cursor: nil)
      Note over Parent: cursor is nil →<br/>user block NOT invoked
      Parent->>Parent: enqueue_jobs(self.class, arguments)
      Parent->>Q: perform_all_later([job_0..job_N-1])
      Note right of Parent: each child gets<br/>cursor_position =<br/>{instance: i, inner_cursor: nil}
      deactivate Parent

      Q->>Child: perform<br/>(cursor: {instance: i, inner_cursor: nil})
      activate Child
      Child->>Child: build_enumerator(cursor: {instance: i, ...})
      Note over Child: cursor non-nil →<br/>user block IS invoked,<br/>builds inner enum for instance i

      loop until done or interrupted
          Child->>Child: each_iteration(record)
          Note right of Child: cursor_position =<br/>{instance: i, inner_cursor: X}
      end

      alt interrupted
          Child->>Q: retry_job (carries current cursor)
          Q->>Child: perform<br/>(cursor: {instance: i, inner_cursor: X})
          Note over Child: cursor still non-nil →<br/>no re-fan-out,<br/>resumes from inner_cursor X
      else completes
          Note over Child: on_complete callback fires
      end
      deactivate Child
```

### Gotchas

- Callback behaviour could be unexpected, as the parent job runs no callbacks, and each child job runs the full set. So `on_start` means this particular instance is starting and `on_complete` means this particular instance is done iterating. There is no "all child jobs are done iterating" callback, as that would require some form of external synchronization.

### Follow-up

- I plan to add parallel versions of most existing enumerators that divvy up the work for you, so you can simply call `enumerator_builder.parallel_active_record_on_records(Product.all, instances: 5, cursor: cursor)`
- This needs documentation

### DIFF
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -141,6 +141,14 @@ module JobIteration
         return
       end
 
+      if enumerator.is_a?(ParallelEnumerator::EnqueueJobs)
+        tags = instrumentation_tags.merge(instances: enumerator.instances)
+        ActiveSupport::Notifications.instrument("enqueue_parallel_jobs.iteration", tags) do
+          enumerator.enqueue_jobs(self)
+        end
+        return
+      end
+
       assert_enumerator!(enumerator)
 
       if executions == 1 && times_interrupted == 0
--- a/lib/job-iteration/parallel_enumerator.rb
+++ b/lib/job-iteration/parallel_enumerator.rb
@@ -1,0 +1,51 @@
+# typed: true
+# frozen_string_literal: true
+
+module JobIteration
+  # ParallelEnumerator allows you to parallelize iterations.
+  class ParallelEnumerator
+    class EnqueueError < StandardError; end
+
+    class EnqueueJobs
+      def initialize(instances)
+        @instances = instances
+      end
+
+      attr_reader :instances
+
+      def enqueue_jobs(job)
+        child_jobs = instances.times.map do |index|
+          job.class.new(*job.arguments).tap do |child_job|
+            child_job.cursor_position = { "instance" => index, "inner_cursor" => nil }
+
+            # Carry forward potential overrides from the parent job
+            child_job.queue_name = job.queue_name
+            child_job.priority = job.priority if job.priority
+          end
+        end
+
+        ActiveJob.perform_all_later(child_jobs)
+
+        unless child_jobs.all?(&:successfully_enqueued?)
+          failed_count = instances - child_jobs.count(&:successfully_enqueued?)
+          raise EnqueueError, "Failed to enqueue #{failed_count} out of #{instances} child jobs"
+        end
+      end
+    end
+
+    def initialize(block, instances:, cursor:)
+      @instance = cursor["instance"]
+      inner_cursor = cursor["inner_cursor"]
+      @inner_enum = block.call(@instance, instances, inner_cursor)
+    end
+
+    def to_enum
+      Enumerator.new(-> { @inner_enum.size }) do |yielder|
+        @inner_enum.each do |object_from_enumerator, cursor_from_enumerator|
+          parallel_cursor = { "instance" => @instance, "inner_cursor" => cursor_from_enumerator }
+          yielder.yield(object_from_enumerator, parallel_cursor)
+        end
+      end
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ nil
 
 ### Features
 
-nil
+- [702](https://github.com/Shopify/job-iteration/pull/702) - Add support for parallel iteration with `enumerator_builder.parallel`. This enqueues multiple jobs, allowing you to split up the work across multiple instances.
 
 ### Bug fixes
 
--- a/dev.yml
+++ b/dev.yml
@@ -11,8 +11,12 @@ up:
   - redis
   - custom:
       name: Create Job Iteration database
-      meet: mysql -uroot -h $MYSQL_HOST -P $MYSQL_PORT -e "CREATE DATABASE job_iteration_test"
-      met?: mysql -uroot -h $MYSQL_HOST -P $MYSQL_PORT job_iteration_test -e "SELECT 1" &> /dev/null
+      meet: |
+        DB_NAME="job_iteration_test${SERVICE_NAMESPACE:+_$SERVICE_NAMESPACE}"
+        mysql -uroot -h $MYSQL_HOST -P $MYSQL_PORT -e "CREATE DATABASE IF NOT EXISTS \`$DB_NAME\`"
+      met?: |
+        DB_NAME="job_iteration_test${SERVICE_NAMESPACE:+_$SERVICE_NAMESPACE}"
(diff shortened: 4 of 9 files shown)


lib/job-iteration/iteration.rb
```diff
@@ -141,6 +141,11 @@ def interruptible_perform(*arguments)
         return
       end
 
+      if enumerator.is_a?(ParallelEnumerator::EnqueueJobs)
+        enumerator.enqueue_jobs(self.class, arguments)
```
akinshonibare: What happens if `perform_all_later` partially fails? this is possible right ? should we be validating this

lib/job-iteration/parallel_enumerator.rb
```diff
@@ -0,0 +1,35 @@
+# typed: true
+# frozen_string_literal: true
+
+module JobIteration
+  # ParallelEnumerator allows you to parallelize iterations.
+  class ParallelEnumerator
+    class EnqueueJobs
+      def initialize(instances)
+        @instances = instances
```
akinshonibare, line 11: Should we also validate that a positive number is passed

lib/job-iteration/parallel_enumerator.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      def enqueue_jobs(job_class, arguments)
+        jobs = @instances.times.map do |index|
+          job_class.new(*arguments).tap { |job| job.cursor_position = { "instance" => index, "inner_cursor" => nil } }
+        end
+        ActiveJob.perform_all_later(jobs)
+      end
+    end
```
MatthewRBruce: Parallel fan-out rebuilds each child as `job_class.new(*arguments)` and only copies `cursor_position`. 

Any instance-level enqueue metadata on the current job is discarded. A realistic path is `MyJob.set(queue: :low, priority: 50).perform_later(...)`: the parent executes with those runtime overrides, but each child is recreated from class defaults, so shards can be enqueued onto the wrong queue or with the wrong priority. 

`Iteration#interruptible_perform` currently passes only `self.class` and `arguments` into `enqueue_jobs`, and `enqueue_jobs` never copies routing metadata from the parent instance. The new tests only verify child class and arguments, so this behavior is uncovered. In production this can misroute time-sensitive or isolated workloads, cause queue starvation, or leak work into default queues.

MatthewRBruce (approved):

lib/job-iteration/parallel_enumerator.rb
Mangara: Fixed. Looks like `queue_name` and `priority` are the only ones that need to be passed through: https://github.com/rails/rails/blob/main/activejob/lib/active_job/core.rb#L175

lib/job-iteration/iteration.rb
Mangara: Yeah, good catch. I added a check that raises a custom error when that happens.

lib/job-iteration/parallel_enumerator.rb
Mangara: Yah, good call. Will add that in the builder, where the other validations happen too.

CelinaAssal (approved):
Nice!

akinshonibare (approved):

lib/job-iteration/iteration.rb
```diff
@@ -141,6 +141,11 @@ def interruptible_perform(*arguments)
         return
       end
 
+      if enumerator.is_a?(ParallelEnumerator::EnqueueJobs)
+        enumerator.enqueue_jobs(self)
```
adrianna-chang-shopify: Is it worth instrumenting the fan-out here?

lib/job-iteration/parallel_enumerator.rb
```diff
@@ -0,0 +1,49 @@
...
+        @instances = instances
+      end
+
+      def enqueue_jobs(job)
+        child_jobs = @instances.times.map do |index|
+          job.class.new(*job.arguments).tap do |child_job|
+            child_job.cursor_position = { "instance" => index, "inner_cursor" => nil }
+
+            # Carry forward potential overrides from the parent job
+            child_job.queue_name = job.queue_name if job.queue_name
```
adrianna-chang-shopify: Active Job provides a default `"default"` queue name, so I think we can drop the guard?
```suggestion
            child_job.queue_name = job.queue_name
```

lib/job-iteration/parallel_enumerator.rb
```diff
@@ -0,0 +1,49 @@
...
+            child_job.priority = job.priority if job.priority
+          end
+        end
+
+        ActiveJob.perform_all_later(child_jobs)
+
+        unless child_jobs.all?(&:successfully_enqueued?)
+          failed_count = @instances - child_jobs.count(&:successfully_enqueued?)
+          raise EnqueueError, "Failed to enqueue #{failed_count} out of #{@instances} child jobs"
```
adrianna-chang-shopify: Hmm, does this mean that if _any_ of the child jobs fail, we re-enqueue all of the children again on retry? I guess we assume idempotency and it's not really an issue?

adrianna-chang-shopify (approved):
Couple small questions but looks good overall!

lib/job-iteration/parallel_enumerator.rb
Mangara: Yeah. This is a tricky case where I don't see any perfect options.

I suspect that usually either all the enqueues will fail, or all will succeed. If we do have a partial success, retrying the job will indeed enqueue additional copies.

lib/job-iteration/iteration.rb
Mangara: Done. We have an extra log line and an event that users can subscribe to:

```
Performing ParallelTestJob (Job ID: a0bf1abe-282c-4af5-84f0-6e6cd9122d8e) from (default) enqueued at 2026-04-30T15:08:10.508344000Z with arguments: {}
Enqueued 2 jobs to REDACTED (2 ParallelTestJob)
[JobIteration::Iteration] Enqueued 2 parallel jobs.
Performed ParallelTestJob (Job ID: a0bf1abe-282c-4af5-84f0-6e6cd9122d8e) from REDACTED(default) in 36.53ms
```

PR #705, opened 2026-05-06
Add parallel array enumerator

Mangara:
### What and why?

This PR adds a parallel array enumerator that builds on the foundation laid in https://github.com/Shopify/job-iteration/pull/702.

While it's not complex, getting the logic exactly right so each instance handles roughly the same number of elements and no elements are double-processed or missed is a bit tricky. That's why it's useful for this to live in the gem, instead of every user having to rebuild it.

### DIFF
--- a/lib/job-iteration/enumerator_builder.rb
+++ b/lib/job-iteration/enumerator_builder.rb
@@ -66,6 +66,23 @@ module JobIteration
       wrap(self, enumerable.each_with_index.drop(drop).to_enum { enumerable.size - drop })
     end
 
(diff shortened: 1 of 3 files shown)


akinshonibare (approved):
LGTM, do we enqueue no-op children, i.e if the instances > array.size or array is empty

Mangara:
> LGTM, do we enqueue no-op children, i.e if the instances > array.size or array is empty

We do. I'm not sure if it's worth guarding against that. A no-op job is pretty harmless.

philippeboyd (approved):

PR #707, opened 2026-05-11
Remove work-arounds for Rails < 7.1

Mangara:
### What and why?

We dropped support for Rails 7.0 in https://github.com/Shopify/job-iteration/pull/704, but still had a bunch of work-arounds in the code. This PR cleans those up.

### DIFF
--- a/lib/job-iteration/active_record_batch_enumerator.rb
+++ b/lib/job-iteration/active_record_batch_enumerator.rb
@@ -66,10 +66,6 @@ module JobIteration
       pkey = @column_mgr.primary_key
       pkey_values = primary_key_values
 
-      # If the primary key is only composed of a single column, simplify the
-      # query. This keeps us compatible with Rails prior to 7.1 where composite
-      # primary keys were introduced along with the syntax that allows you to
-      # query for multi-column values.
       if pkey.size <= 1
         pkey = pkey.first
         pkey_values = pkey_values.map(&:first)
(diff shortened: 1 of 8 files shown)


SomalianIvan (approved):

lib/job-iteration/active_record_batch_enumerator.rb
```diff
@@ -66,10 +66,6 @@ def filter_relation_with_primary_key(primary_key_values)
       pkey = @column_mgr.primary_key
       pkey_values = primary_key_values
 
-      # If the primary key is only composed of a single column, simplify the
-      # query. This keeps us compatible with Rails prior to 7.1 where composite
-      # primary keys were introduced along with the syntax that allows you to
-      # query for multi-column values.
```
beepubapu, line 72: Should we leave a comment here clarifying why the branch exists, if not for < 7.1 support?

beepubapu (approved):
One clarifying question about a remaining code branch.

lib/job-iteration/active_record_batch_enumerator.rb
Mangara: Simplifying the query still seems useful to me. Does that require a comment?

beepubapu: Nope, nit. Move ahead.

Issue #714, opened 2026-05-25
Support keyword arguments in `build_enumerator`


SomalianIvan:
`job-iteration` currently supports enqueuing jobs with keyword arguments, but those arguments are passed to `build_enumerator` as a positional Hash.

For example, this works today:

```ruby
class ExampleJob
  def build_enumerator(params, cursor:)
    account_id = params.fetch(:account_id)
    ...
  end
  ...
end

ExampleJob.perform_later(account_id: 123)
```

But this does not work:

```ruby
class ExampleJob < ActiveJob::Base
  include JobIteration::Iteration

  def build_enumerator(account_id:, cursor:)
    ...
  end
ExampleJob.perform_later(account_id: 123)
```

The guide currently documents this limitation:

> Note that you cannot use `ruby2_keywords` at present, and the keyword arguments syntax is not supported in `each_iteration` / `build_enumerator`.

### Motivation

This would improve a few things:

1. **Usability**

   Accessing everything through a `params` Hash is less ergonomic:

   ```ruby
   account_id = params.fetch(:account_id)
   locale = params.fetch(:locale)
   dry_run = params.fetch(:dry_run, false)
   ```

   Keyword arguments make the job contract easier to read at the method boundary:

   ```ruby
   def build_enumerator(account_id:, locale:, dry_run: false, cursor:)
   ```


2. **Better typing support**

   Static typing and RBI-generation tooling can generally model required and optional/defaulted keyword arguments more naturally than optional key/value pairs inside a Hash shape. [sorbet issue](https://github.com/sorbet/sorbet/issues/10303)

3. **Fewer argument-normalization shims**

   Code using Hash params often grows compatibility helpers such as indifferent-access conversion or symbol/string key fallbacks. Supporting kwargs would let new jobs avoid that pattern.


### Proposed behavior

Support keyword argument signatures when the target method declares keyword parameters.

### Backwards compatibility

Existing jobs that accept keyword arguments as a positional Hash should continue to work unchanged:

```ruby
def build_enumerator(params, cursor:)
  # existing behavior
end

def each_iteration(record, params)
  # existing behavior
end
```

### Open questions

-  Worth considering if we can drop `cursor` from the signature. This would avoid making `cursor:` a reserved keyword.
```ruby                                                                                                                                                                                                                                                                      
   def build_enumerator(account_id:)                                                                                                                                                                                                                                          
     cursor = cursor_position                                                                                                                                                                                                                                                                                                                                                                                          
[...]

PR #715, opened 2026-05-29
Support keyword arguments

SomalianIvan:
## iteration jobs today

Support only one signature shape, a positional params hash and `cursor`:

```ruby
class IterationJob < ActiveJob::Base

    shop_id = params.fetch(:shop_id)
    # ...
  end
end

# all enqueue styles work
IterationJob.perform_later(shop_id: 1) # ruby2_keywords
IterationJob.perform_later({ shop_id: 1 }) # single positional hash
params = {} # some hash we can't statically validate :scary:
IterationJob.perform_later(params) # single positional hash
```

We want to support kwargs, more details on why in #714:

```ruby
class IterationKwargsJob < ActiveJob::Base

  def build_enumerator(shop_id:, cursor:)
    # ...
  end
end

# only the ruby2_keywords enqueue style should work for kwargs signatures
IterationKwargsJob.perform_later(shop_id: 1) # ruby2_keywords

# should not be allowed for kwargs signatures
IterationKwargsJob.perform_later({ shop_id: 1 }) # single positional hash
params = {} # some hash we can't statically validate :scary:
IterationKwargsJob.perform_later(params) # single positional hash
```

### Implementation

- Uses Ruby argument forwarding for the public `perform(...)` wrapper.
- Carries forwarded Active Job keyword arguments explicitly through the iteration dispatch path instead of relying on `Hash.ruby2_keywords_hash` checks.
- Routes `build_enumerator` and `each_iteration` calls through dispatch helpers instead of directly splatting arguments.
- The dispatch helpers inspect the target method signature:
  - if the target method declares job keyword arguments, forwarded kwargs are passed with `**kwargs`;
  - if the target method does not declare keyword arguments, the Hash is left as a positional params argument for backwards compatibility.
- Rejects `cursor:` as a user-provided keyword argument for keyword-aware `build_enumerator` methods because `cursor:` is reserved by job-iteration.
- Updates the argument semantics guide and changelog to document kwargs support and the transitional params-hash compatibility path.

### Compatibility note

Existing jobs using `def build_enumerator(params, cursor:)` with `perform_later(keyword: value)` continue to work for now. That compatibility path is transitional; jobs should migrate to explicit keyword signatures before a future major release removes positional params-hash support for keyword-style enqueues.

### DIFF
--- a/guides/argument-semantics.md
+++ b/guides/argument-semantics.md
@@ -56,21 +56,17 @@ ArgumentativeJob.perform_later(_arg1 = "One", _arg2 = "Two", _arg3 = "Three")
 
 ### Jobs with keyword arguments
 
-Jobs with keyword arguments will have the keyword arguments available to both `build_enumerator` and `each_iteration`, but these arguments come packaged into a Hash in both cases. You will need to `fetch` or `[]` your parameter from the `Hash` you get passed in:
+Jobs with keyword arguments can declare those keyword arguments directly on both `build_enumerator` and `each_iteration`:
 
 ```ruby
 class ParameterizedJob < ActiveJob::Base
   include JobIteration::Iteration
 
-  def build_enumerator(kwargs, cursor:)
-    name = kwargs.fetch(:name)
-    email = kwargs.fetch(:email)
+  def build_enumerator(name:, email:, cursor:)
     # ...
   end
 
-  def each_iteration(object_yielded_from_enumerator, kwargs)
-    name = kwargs.fetch(:name)
-    email = kwargs.fetch(:email)
+  def each_iteration(object_yielded_from_enumerator, name:, email:)
     # ...
   end
 end
@@ -82,37 +78,27 @@ To enqueue the job:
 ParameterizedJob.perform_later(name: "Jane", email: "jane@host.example")
 ```
 
-Note that you cannot use `ruby2_keywords` at present, and the keyword arguments syntax is not supported in `each_iteration` / `build_enumerator`.
+The `cursor:` keyword argument on `build_enumerator` is reserved for `job-iteration`; do not pass a job argument named `cursor` when using keyword arguments.
 
-### Jobs with both positional and keyword arguments
-
-Jobs with keyword arguments will have the keyword arguments available to both `build_enumerator` and `each_iteration`, but these arguments come packaged into a Hash in both cases. You will need to `fetch` or `[]` your parameter from the `Hash` you get passed in. Positional arguments get passed first and "unsplatted" (not combined into an array), the `Hash` containing keyword arguments comes after:
+For compatibility with existing jobs, keyword arguments are still passed as a positional params Hash after any positional arguments when `build_enumerator` and `each_iteration` do not declare keyword arguments (other than `cursor` for `build_enumerator`). This compatibility path is transitional and will be removed in a future release; prefer declaring keyword arguments directly instead of combining a positional params Hash with `perform_later(keyword: value)`:
 
 ```ruby
-class HighlyConfigurableGreetingJob < ActiveJob::Base
+class LegacyParameterizedJob < ActiveJob::Base
   include JobIteration::Iteration
 
-  def build_enumerator(subject_line, kwargs, cursor:)
-    name = kwargs.fetch(:sender_name)
-    email = kwargs.fetch(:sender_email)
+  def build_enumerator(params, cursor:)
+    name = params.fetch(:name)
+    email = params.fetch(:email)
     # ...
   end
 
-  def each_iteration(object_yielded_from_enumerator, subject_line, kwargs)
-    name = kwargs.fetch(:sender_name)
-    email = kwargs.fetch(:sender_email)
+  def each_iteration(object_yielded_from_enumerator, params)
+    name = params.fetch(:name)
+    email = params.fetch(:email)
     # ...
   end
 end
 ```
-
-To enqueue the job:
-
-```ruby
-HighlyConfigurableGreetingJob.perform_later(_subject_line = "Greetings everybody!", sender_name: "Jane", sender_email: "jane@host.example")
-```
-
-Note that you cannot use `ruby2_keywords` at present, and the keyword arguments syntax is not supported in `each_iteration` / `build_enumerator`.
 
 ### Returning (yielding) from enumerators
 
--- a/lib/job-iteration/iteration.rb
+++ b/lib/job-iteration/iteration.rb
@@ -107,8 +107,8 @@ module JobIteration
       self.total_time = Float(job_data["total_time"] || 0.0)
     end
 
-    def perform(*params) # @private
-      interruptible_perform(*params)
+    def perform(...) # @private
+      interruptible_perform(...)
 
       nil
     end
@@ -128,12 +128,12 @@ module JobIteration
       JobIteration.enumerator_builder.new(self)
     end
 
-    def interruptible_perform(*arguments)
+    def interruptible_perform(*args, **kwargs)
       self.start_time = Time.now.utc
 
       enumerator = nil
       ActiveSupport::Notifications.instrument("build_enumerator.iteration", instrumentation_tags) do
-        enumerator = build_enumerator(*arguments, cursor: cursor_position)
+        enumerator = call_job_iteration_build_enumerator(args, kwargs)
       end
 
       unless enumerator
@@ -161,7 +161,7 @@ module JobIteration
       end
 
       completed = catch(:abort) do
-        iterate_with_enumerator(enumerator, arguments)
+        iterate_with_enumerator(enumerator, args, kwargs)
       end
 
       run_callbacks(:shutdown)
@@ -178,8 +178,7 @@ module JobIteration
       end
     end
 
-    def iterate_with_enumerator(enumerator, arguments)
-      arguments = arguments.dup.freeze
+    def iterate_with_enumerator(enumerator, args, kwargs)
       found_record = false
       @needs_reenqueue = false
 
@@ -191,7 +190,7 @@ module JobIteration
         ActiveSupport::Notifications.instrument("each_iteration.iteration", tags) do
           found_record = true
           run_callbacks(:iterate) do
-            each_iteration(object_from_enumerator, *arguments)
+            call_job_iteration_each_iteration(object_from_enumerator, args, kwargs)
           end
           self.cursor_position = cursor_from_enumerator
         end
@@ -213,6 +212,64 @@ module JobIteration
       true
     ensure
       adjust_total_time
+    end
+
+    def call_job_iteration_build_enumerator(args, kwargs)
+      positional_args, keyword_args = normalize_job_iteration_arguments(args, kwargs, :build_enumerator)
+
+      if keyword_args&.key?(:cursor)
+        raise ArgumentError, "The keyword argument `cursor` is reserved for the job iteration framework. " \
+          "Please remove `cursor` from the arguments passed to the job or rename it"
+      end
+
+      # `keyword_args || {}` because splatting `nil` raises on Ruby < 3.4; it only
+      # became a no-op in 3.4 (https://bugs.ruby-lang.org/issues/20064).
+      build_enumerator(*positional_args, **(keyword_args || {}), cursor: cursor_position)
+    end
+
+    def call_job_iteration_each_iteration(object_from_enumerator, args, kwargs)
+      positional_args, keyword_args = normalize_job_iteration_arguments(args, kwargs, :each_iteration)
+      # `keyword_args || {}` because splatting `nil` raises on Ruby < 3.4; it only
+      # became a no-op in 3.4 (https://bugs.ruby-lang.org/issues/20064).
+      each_iteration(object_from_enumerator, *positional_args, **(keyword_args || {}))
+    end
+
+    # Normalize Active Job kwargs for job-iteration dispatch. If the target method
+    # accepts job keyword parameters other than build_enumerator's reserved cursor:,
+    # keep kwargs separate so they can be splatted. Otherwise, append kwargs as a
+    # positional params Hash for transitional compatibility with existing jobs
+    # enqueued with keyword syntax.
+    #: (Array[top], Hash[Symbol, top], Symbol) -> [Array[top], Hash[Symbol, top]?]
+    def normalize_job_iteration_arguments(args, kwargs, method_name)
+      if kwargs.empty?
+        [args.dup.freeze, nil]
+      elsif job_has_keyword_parameters?(method_name)
+        [args.dup.freeze, kwargs.dup.freeze]
+      else
+        params_hash_args = args + [kwargs]
+        [params_hash_args.dup.freeze, nil]
+      end
+    end
+
+    #: (Symbol) -> bool
+    def job_has_keyword_parameters?(method_name)
+      @job_has_keyword_parameters ||= {}
+      return @job_has_keyword_parameters[method_name] if @job_has_keyword_parameters.key?(method_name)
+
+      @job_has_keyword_parameters[method_name] = method_parameters(method_name).any? do |type, name|
+        job_keyword_argument?(method_name, type, name)
+      end
+    end
+
+    def job_keyword_argument?(method_name, type, name)
+      # Match keyword parameters: double-splat (**kwargs), required (key:) or optional (key: default).
+      return false unless type == :keyrest || type == :keyreq || type == :key
+
+      # Ignore the `cursor:` argument, which is part of the `job-iteration`
+      # framework, and not a argument in the job's serialized argument list.
+      return false if method_name == :build_enumerator && name == :cursor
+
+      true
     end
 
     def reenqueue_iteration_job
--- a/test/unit/active_job_iteration_test.rb
+++ b/test/unit/active_job_iteration_test.rb
@@ -48,6 +48,122 @@ module JobIteration
       end
     end
 
+    class KwargsIterationJob < SimpleIterationJob
+      def build_enumerator(one_arg:, another_arg:, cursor:)
+        enumerator_builder.build_times_enumerator(2, cursor: cursor)
+      end
+
+      def each_iteration(item, one_arg:, another_arg:)
+        self.class.records_performed << [item, one_arg, another_arg]
+      end
+    end
+
+    class MixedArgumentsIterationJob < SimpleIterationJob
+      def build_enumerator(positional_arg, kwarg1:, kwarg2:, cursor:)
+        enumerator_builder.build_times_enumerator(2, cursor: cursor)
+      end
+
+      def each_iteration(item, positional_arg, kwarg1:, kwarg2:)
+        self.class.records_performed << [item, positional_arg, kwarg1, kwarg2]
+      end
+    end
+
+    class KeyrestIterationJob < SimpleIterationJob
+      def build_enumerator(cursor:, **_kwargs)
+        enumerator_builder.build_times_enumerator(2, cursor: cursor)
+      end
+
+      def each_iteration(item, **kwargs)
+        self.class.records_performed << [item, kwargs]
+      end
+    end
+
+    class PositionalOnlyIterationJob < SimpleIterationJob
+      def build_enumerator(cursor:)
+        enumerator_builder.build_times_enumerator(1, cursor: cursor)
+      end
+
+      def each_iteration(item)
+        self.class.records_performed << item
+      end
+    end
+
+    def test_supports_keyword_arguments
+      KwargsIterationJob.perform_later(one_arg: "foo", another_arg: "bar")
+      work_one_job
+      expected = [
+        [0, "foo", "bar"],
+        [1, "foo", "bar"],
+      ]
+      assert_equal(expected, KwargsIterationJob.records_performed)
+    end
+
+    def test_supports_mixed_arguments
+      MixedArgumentsIterationJob.perform_later("foo", kwarg1: "bar", kwarg2: "baz")
+      work_one_job
+      expected = [
+        [0, "foo", "bar", "baz"],
+        [1, "foo", "bar", "baz"],
+      ]
+      assert_equal(expected, MixedArgumentsIterationJob.records_performed)
+    end
+
+    def test_supports_keyword_arguments_after_interruption
+      iterate_exact_times(1.times)
+      KwargsIterationJob.perform_later(one_arg: "foo", another_arg: "bar")
+
+      work_one_job
+      assert_equal([[0, "foo", "bar"]], KwargsIterationJob.records_performed)
+
+      work_one_job
+      expected = [
+        [0, "foo", "bar"],
+        [1, "foo", "bar"],
+      ]
+      assert_equal(expected, KwargsIterationJob.records_performed)
+    end
+
+    def test_supports_mixed_arguments_after_interruption
+      iterate_exact_times(1.times)
+      MixedArgumentsIterationJob.perform_later("foo", kwarg1: "bar", kwarg2: "baz")
+
+      work_one_job
+      assert_equal([[0, "foo", "bar", "baz"]], MixedArgumentsIterationJob.records_performed)
+
+      work_one_job
+      expected = [
+        [0, "foo", "bar", "baz"],
+        [1, "foo", "bar", "baz"],
+      ]
+      assert_equal(expected, MixedArgumentsIterationJob.records_performed)
+    end
+
+    def test_supports_keyrest_arguments
+      KeyrestIterationJob.perform_later(one_arg: "foo", another_arg: "bar")
+      work_one_job
+      expected = [
+        [0, { one_arg: "foo", another_arg: "bar" }],
+        [1, { one_arg: "foo", another_arg: "bar" }],
+      ]
+      assert_equal(expected, KeyrestIterationJob.records_performed)
+    end
+
+    def test_supports_positional_only_each_iteration_without_keywords
+      PositionalOnlyIterationJob.perform_later
+      work_one_job
+      assert_equal([0], PositionalOnlyIterationJob.records_performed)
+    end
+
+    def test_rejects_reserved_cursor_keyword_argument
+      KwargsIterationJob.perform_later(one_arg: "foo", another_arg: "bar", cursor: "custom-cursor")
+
+      error = assert_raises(ArgumentError) do
+        work_one_job
+      end
+
+      assert_match(/keyword argument `cursor` is reserved/, error.message)
+    end
+
     class ParamsIterationJob < SimpleIterationJob
       def build_enumerator(params, cursor:)
         enumerator_builder.build_times_enumerator(params.fetch(:times, 2), cursor: cursor)
@@ -601,6 +717,13 @@ module JobIteration
     def test_passes_params_to_each_iteration
       params = { "walrus" => "best" }
       push(ParamsIterationJob, params)
+      work_one_job
+      assert_equal([params, params], ParamsIterationJob.records_performed)
+    end
+
+    def test_keeps_keyword_arguments_as_params_hash_when_methods_do_not_accept_keywords
+      params = { times: 2 }
+      ParamsIterationJob.perform_later(**params)
       work_one_job
       assert_equal([params, params], ParamsIterationJob.records_performed)
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ nil
 
 ### Features
 
-nil
+- Add support for keyword arguments in `build_enumerator` and `each_iteration` while preserving positional params Hash compatibility for existing jobs. This compatibility path is transitional; migrate jobs away from positional params Hash signatures paired with `perform_later(keyword: value)`.
 
 ### Bug fixes
 


lib/job-iteration/iteration.rb
```diff
@@ -112,6 +112,7 @@ def perform(*params) # @private
 
       nil
     end
+    ruby2_keywords(:perform) if respond_to?(:ruby2_keywords, true)
```
rafaelfranca: We can probably drop support to Ruby versions that don't support the keyword arguments and the we can stop caring about calling this method. Only Ruby 3 are supported today

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +218,49 @@ def iterate_with_enumerator(enumerator, arguments)
...
+    def ruby2_keywords_hash?(argument)
+      argument.is_a?(Hash) && Hash.ruby2_keywords_hash?(argument)
+    end
+
+    def accepts_job_keyword_arguments?(method_name)
+      @accepts_job_keyword_arguments ||= {}
+      return @accepts_job_keyword_arguments[method_name] if @accepts_job_keyword_arguments.key?(method_name)
+
+      @accepts_job_keyword_arguments[method_name] = method_parameters(method_name).any? do |type, name|
+        type == :keyrest || job_keyword_argument?(method_name, type, name)
```
amomchilov: `job_keyword_argument?` already checks for `type == :keyrest`


```suggestion
        job_keyword_argument?(method_name, type, name)
```

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +218,49 @@ def iterate_with_enumerator(enumerator, arguments)
...
+      if keyword_arguments
+        each_iteration(object_from_enumerator, *positional_arguments, **keyword_arguments)
+      else
+        each_iteration(object_from_enumerator, *arguments)
+      end
+    end
+
+    def extract_keyword_arguments(arguments, method_name)
+      if accepts_job_keyword_arguments?(method_name) && ruby2_keywords_hash?(arguments.last)
+        [arguments[0...-1], arguments.last]
```
amomchilov: Document the intent:

```suggestion
        # Split off the keyword args Hash, so it can later be splatted into the call to `each_iteration`
        *positional_args, kwargs_hash = arguments
        [positional_args, kwargs_hash]
```

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +218,49 @@ def iterate_with_enumerator(enumerator, arguments)
...
+    def call_each_iteration(object_from_enumerator, arguments)
+      positional_arguments, keyword_arguments = extract_keyword_arguments(arguments, :each_iteration)
+      if keyword_arguments
+        each_iteration(object_from_enumerator, *positional_arguments, **keyword_arguments)
+      else
+        each_iteration(object_from_enumerator, *arguments)
+      end
+    end
+
+    def extract_keyword_arguments(arguments, method_name)
```
amomchilov: This is some tricky stuff, I think some documentation on these methods could really help.

Especially adding RBS types. I know this repo doesn't have Sorbet set up (yet? :) ), but it's still helpful documentation.


```suggestion
    # Check if the given `method_name`'s declaration accepts keyword arguments,
    # and split up the given `arguments` Array to extract out the Kwargs hash, if any.
    #: (Array[top], Symbol) -> [Array[top], top?]
    def extract_keyword_arguments(arguments, method_name)
```

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +218,49 @@ def iterate_with_enumerator(enumerator, arguments)
...
+      return @accepts_job_keyword_arguments[method_name] if @accepts_job_keyword_arguments.key?(method_name)
+
+      @accepts_job_keyword_arguments[method_name] = method_parameters(method_name).any? do |type, name|
+        type == :keyrest || job_keyword_argument?(method_name, type, name)
+      end
+    end
+
+    def job_keyword_argument?(method_name, type, name)
+      [:key, :keyreq].include?(type) && !(method_name == :build_enumerator && name == :cursor)
+    end
```
amomchilov, line 273:
```suggestion
    def job_keyword_argument?(method_name, type, name)
      # Only look for keyword args (required or optional)
      return false unless type == keyreq || type == :key
      
      # Ignore the `cursor:` argument, which is part of the `job-iteration`
      # framework, and not a argument in the job's serialized argument list.
      return false if method_name == :build_enumerator && name == :cursor
    end
```

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +218,49 @@ def iterate_with_enumerator(enumerator, arguments)
...
+      else
+        [arguments, nil]
+      end
+    end
+
+    def ruby2_keywords_hash?(argument)
+      argument.is_a?(Hash) && Hash.ruby2_keywords_hash?(argument)
+    end
+
+    def accepts_job_keyword_arguments?(method_name)
```
amomchilov: I think we should rename most usages of "arguments" around here, because we're actually checking the _parameters_ of the given method, rather than the _arguments_ from one particular call to the method. 

E.g. the `@accepts_job_keyword_arguments[...] = method_parameters(...)` line reads funny (to me, at least). The left side names after arguments, but the right side is talking about parameters


```suggestion
    #: (Symbol) -> bool
    def has_job_keyword_parameters?(method_name)
```

test/unit/active_job_iteration_test.rb
```diff
@@ -48,6 +48,46 @@ def each_iteration(item, one_arg, another_arg)
...
+      end
+    end
+
+    class MixedArgumentsIterationJob < SimpleIterationJob
+      def build_enumerator(positional_arg, keyword_arg:, cursor:)
+        enumerator_builder.build_times_enumerator(2, cursor: cursor)
+      end
+
+      def each_iteration(item, positional_arg, keyword_arg:)
+        self.class.records_performed << [item, positional_arg, keyword_arg]
```
amomchilov: let's do multiple just in case


```suggestion
      def each_iteration(item, positional_arg, kwarg1:, kwarg2:)
        self.class.records_performed << [item, positional_arg, kwarg1, kwarg2]
```

Q: Should we support `**remaining_kwargs`?

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +218,49 @@ def iterate_with_enumerator(enumerator, arguments)
...
+      end
+    end
+
+    def call_each_iteration(object_from_enumerator, arguments)
+      positional_arguments, keyword_arguments = extract_keyword_arguments(arguments, :each_iteration)
+      if keyword_arguments
+        each_iteration(object_from_enumerator, *positional_arguments, **keyword_arguments)
+      else
+        each_iteration(object_from_enumerator, *arguments)
+      end
```
amomchilov: One alternative I considered is that you can actually make this unconditional (because `nil` is safe to splat, like like `**nil`):


```suggestion
      each_iteration(object_from_enumerator, *positional_arguments, **keyword_arguments)
```

But I think your explicit check is more clear, and better highlights the two distinct cases we care about.

amomchilov (commented):
I think this makes sense. The tests give me some confidence, but I would like someone from Rails infra to doubly check over everything.

lib/job-iteration/iteration.rb
SomalianIvan: I thought it was necessary if we want to differentiate between callsite(enqueue) variants

```ruby
def perform(*args)
  puts("#{Hash.ruby2_keywords_hash?(args.last)}")
end
ruby2_keywords(:perform)

perform(shop_id: 123) # ruby2_keywords_hash? => true
perform({ shop_id: 123 }) # ruby2_keywords_hash? => false
```

but if I understand your suggestion we should double splat kwargs:

```ruby
def perform(*args,**kwargs)
  # if build_enumerator accepts kwargs pass kwargs
  # if build_enumerator accepts (params, cursor:) maintain current behaviour
end
```

is that what you mean?

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +218,55 @@ def iterate_with_enumerator(enumerator, arguments)
...
+    end
+
+    def job_keyword_argument?(method_name, type, name)
+      # Only look for keyword args (required or optional)
+      return false unless type == keyreq || type == :key
+      
+      # Ignore the `cursor:` argument, which is part of the `job-iteration`
+      # framework, and not a argument in the job's serialized argument list.
+      return false if method_name == :build_enumerator && name == :cursor
```
amomchilov, line 270: Oops I forgot the `true` at the end


```suggestion
      return false if method_name == :build_enumerator && name == :cursor
      
      true
```

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +218,58 @@ def iterate_with_enumerator(enumerator, arguments)
       adjust_total_time
     end
 
+    def call_build_enumerator(arguments)
+      positional_arguments, keyword_arguments = extract_keyword_arguments(arguments, :build_enumerator)
+      if keyword_arguments
+        build_enumerator(*positional_arguments, **keyword_arguments, cursor: cursor_position)
```
paracycle: What happens if `keyword_arguments` includes the key `:cursor`? As far as I can see, you get a warning, but silently drop the `cursor:` value in the kwargs. We might want to explicitly guard against that case and raise an exception early if anyone passes a `cursor:` keyword.
```ruby
irb> def foo(*args, **kwargs) = puts(args, kwargs)
#=> :foo

irb> foo(**{a: 1, b: 2}, b: 3)
(irb):10: warning: key :b is duplicated and overwritten on line 10
{a: 1, b: 3}
#=> nil
```

lib/job-iteration/iteration.rb
paracycle: Yeah, I think we should move to:
```ruby
def perform(*args, **kwargs)
```
and drop `ruby2_keywords` support.

IMO, `perform({ shop_id: 123 })` is ambiguous. You can't be sure if the user meant to pass a hash as the only positional argument, or if they are trying to pass keyword arguments.

Considering that we don't have support for keyword arguments today, I think we can restrict it so that these calls are fail. Otherwise, we might have cases where the caller tried to pass `{ foo: 42 }` as the only parameter for the job, not being aware that the job added a `foo:` keyword, and end up in the wrong state.

Additionally, this makes the rest of the code much simpler. You don't have to check that the target method declares kwargs or not, you just pass them on as args and kwargs, and the method call fails if they don't match what's expected.

lib/job-iteration/iteration.rb
SomalianIvan: good catch, I'll add a guard against that.

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +219,65 @@ def iterate_with_enumerator(enumerator, arguments)
       adjust_total_time
     end
 
+    def call_build_enumerator(arguments, forwarded_keyword_arguments)
```
paracycle: I still don't understand why this method isn't just passing all the arguments to `build_enumerator`? That way if `build_enumerator` accepts kwargs, then the call will only work if the user has supplied kwargs, otherwise it will fail without us having to do any method signature checks.

Moreover, this will make writing the Tapioca DSL compiler much harder as well, since now the signature of `perform` is being decoupled from the signature of `build_enumerator`

lib/job-iteration/iteration.rb
```diff
@@ -128,12 +128,16 @@ def enumerator_builder
       JobIteration.enumerator_builder.new(self)
     end
 
-    def interruptible_perform(*arguments)
+    def interruptible_perform(*arguments, **kwargs)
+      keyword_arguments = kwargs unless kwargs.empty?
+      arguments = arguments.dup
+      arguments << kwargs if keyword_arguments
```
paracycle: Why is this needed?

lib/job-iteration/iteration.rb
SomalianIvan: This is for backwards compatibility with existing params-hash jobs. With this method signature, Active Job kwargs are split out of `arguments`, so without appending them back a legacy job would stop receiving its params:

```ruby
class LegacyJob < ActiveJob::Base

    params.fetch(:shop_id)
  end

  def each_iteration(record, params)
    # ...
  end
end

LegacyJob.perform_later(shop_id: 1)
```

For that job shape we still need to call:

```ruby
build_enumerator({ shop_id: 1 }, cursor: cursor)
each_iteration(record, { shop_id: 1 })
```

Keyword-aware jobs use the separate `keyword_arguments` path instead:

```ruby
def build_enumerator(shop_id:, cursor:)
def each_iteration(record, shop_id:)
```

so they receive `**kwargs` rather than the appended positional hash.

lib/job-iteration/iteration.rb
SomalianIvan: https://github.com/Shopify/job-iteration/pull/715#discussion_r3350733114

guides/argument-semantics.md
```diff
@@ -82,25 +78,41 @@ To enqueue the job:
 ParameterizedJob.perform_later(name: "Jane", email: "jane@host.example")
 ```
 
-Note that you cannot use `ruby2_keywords` at present, and the keyword arguments syntax is not supported in `each_iteration` / `build_enumerator`.
+The `cursor:` keyword argument on `build_enumerator` is reserved for `job-iteration`; do not pass a job argument named `cursor` when using keyword arguments.
+
+For compatibility with existing jobs, keyword arguments are still passed as a positional params Hash when `build_enumerator` and `each_iteration` do not declare keyword arguments. This compatibility path is transitional and will be removed in a future release; prefer declaring keyword arguments directly instead of combining a positional params Hash with `perform_later(keyword: value)`:
```
liamstewart:
```suggestion
For compatibility with existing jobs, keyword arguments are still passed as a positional params Hash when `build_enumerator` and `each_iteration` do not declare keyword arguments (other than `cursor` for `build_enumerator`). This compatibility path is transitional and will be removed in a future release; prefer declaring keyword arguments directly instead of combining a positional params Hash with `perform_later(keyword: value)`:
```

guides/argument-semantics.md
```diff
@@ -82,25 +78,41 @@ To enqueue the job:
...
+    email = params.fetch(:email)
+    # ...
+  end
+end
+```
 
 ### Jobs with both positional and keyword arguments
 
-Jobs with keyword arguments will have the keyword arguments available to both `build_enumerator` and `each_iteration`, but these arguments come packaged into a Hash in both cases. You will need to `fetch` or `[]` your parameter from the `Hash` you get passed in. Positional arguments get passed first and "unsplatted" (not combined into an array), the `Hash` containing keyword arguments comes after:
+Jobs with both positional and keyword arguments will have the positional arguments available first, followed by the keyword arguments declared on both `build_enumerator` and `each_iteration`:
```
liamstewart: Can this section go away entirely? Or at least say that there's nothing special about using both positional and keyword args. Calling is positional first, kwargs second.

lib/job-iteration/iteration.rb
liamstewart:
```suggestion
      arguments << kwargs unless kwargs.empty?
```

and use `kwargs` instead of `keyword_arguments` below. Then `extract_keyword_arguments` is

```ruby
def extract_keyword_arguments(arguments, method_name, forwarded_keyword_arguments)
  if has_job_keyword_parameters?(method_name) && !forwarded_keyword_arguments.empty?
    [arguments[0...-1], forwarded_keyword_arguments]
  else
    [arguments, nil]
  end
end
```

lib/job-iteration/iteration.rb
paracycle: What I am failing to understand is this:

Given I have a job defined like this today:
```ruby
class LegacyJob < ActiveJob::Base

    params.fetch(:shop_id)
  end

  def each_iteration(record, params)
    # ...
  end
end
```
and its callers are calling it like this:
```ruby
LegacyJob.perform_later({ shop_id: 1 })
```

After this PR, if I change my job to be:
```ruby
class LegacyJob < ActiveJob::Base

  def build_enumerator(shop_id:, cursor:)
    shop_id
  end

  def each_iteration(record, shop_id:)
    # ...
  end
end
```
then I would also have to change all the usages of the job to be:
```ruby
LegacyJob.perform_later(shop_id: 1)
```

Without mixing positional and keyword arguments, this will work as expected and we won't have to do any method signature introspection to make it work.

What am I missing?

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +219,65 @@ def iterate_with_enumerator(enumerator, arguments)
       adjust_total_time
     end
 
+    def call_build_enumerator(arguments, forwarded_keyword_arguments)
+      positional_arguments, keyword_arguments = extract_keyword_arguments(
+        arguments,
+        :build_enumerator,
+        forwarded_keyword_arguments,
+      )
+      return build_enumerator(*arguments, cursor: cursor_position) unless keyword_arguments
```
liamstewart:
```suggestion
      return build_enumerator(*positional_arguments, cursor: cursor_position) unless keyword_arguments
```

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +219,65 @@ def iterate_with_enumerator(enumerator, arguments)
...
+
+      if keyword_arguments.key?(:cursor)
+        raise ArgumentError, "The keyword argument `cursor` is reserved for the job iteration framework. " \
+          "Please remove `cursor` from the arguments passed to the job or rename"
```
liamstewart:
```suggestion
          "Please remove `cursor` from the arguments passed to the job or rename it"
```

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +219,65 @@ def iterate_with_enumerator(enumerator, arguments)
...
+      @accepts_job_keyword_arguments ||= {}
+      return @accepts_job_keyword_arguments[method_name] if @accepts_job_keyword_arguments.key?(method_name)
+
+      @accepts_job_keyword_arguments[method_name] = method_parameters(method_name).any? do |type, name|
+        job_keyword_argument?(method_name, type, name)
+      end
+    end
+
+    def job_keyword_argument?(method_name, type, name)
+      # look for keyword args (double-splat  **kwargs or required or optional or )
```
liamstewart:
```suggestion
      # Match keyword parameters: double-splat (**kwargs), required (key:) or optional (key: default).
```

test/unit/active_job_iteration_test.rb
```diff
@@ -48,6 +48,92 @@ def each_iteration(item, one_arg, another_arg)
...
+      end
+
+      def each_iteration(item)
+        self.class.records_performed << item
+      end
+    end
+
+    def test_supports_keyword_arguments
+      KwargsIterationJob.perform_later(one_arg: "foo", another_arg: "bar")
+      work_one_job
```
liamstewart, line 93: Would be worth testing interruption  cases for KwargsIterationJob / MixedArgumentsIterationJob similar to `test_passes_params_to_each_iteration_without_extra_information_on_interruption`

lib/job-iteration/iteration.rb
SomalianIvan: we are aligned here - I first wanted to preserve backwards compatibility. There is a note in the PR description:

> Existing jobs using def build_enumerator(params, cursor:) with perform_later(keyword: value) continue to work for now. That compatibility path is transitional; jobs should migrate to explicit keyword signatures before a future major release removes positional params-hash support for keyword-style enqueues.

Do you think we should skip the backwards compatibility and introduce a breaking change?

paracycle: Ah, I see what you are saying now. It is not about the `perform_later({ shop_id: 1 })` calls, it is about `perform_later(shop_id: 1)` calls which work today with jobs that only accept positional args, and might already exist in the codebase, and which will break if we do the breaking change.

Thanks, I had not considered that case, and that was what I was missing.

I agree that we should keep the transitional code. Thanks for being patient with me.

SomalianIvan: > Thanks for being patient with me.

 Likewise thank you for your patience while reviewing this!

lib/job-iteration/iteration.rb
paracycle: It is customary to name these `args` and `kwargs`, and I think we shouldn't be doing the data munging here, but in `extract_keyword_arguments` call.

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +219,65 @@ def iterate_with_enumerator(enumerator, arguments)
...
+    def extract_keyword_arguments(arguments, method_name, forwarded_keyword_arguments)
+      if has_job_keyword_parameters?(method_name) && forwarded_keyword_arguments
+        [arguments[0...-1], forwarded_keyword_arguments]
+      else
+        [arguments, nil]
+      end
+    end
+
+    #: (Symbol) -> bool
+    def has_job_keyword_parameters?(method_name)
```
paracycle: I think:
```suggestion
    def job_has_keyword_parameters?(method_name)
```
is a much better name for this method

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +219,65 @@ def iterate_with_enumerator(enumerator, arguments)
...
+      )
+      return each_iteration(object_from_enumerator, *positional_arguments) unless keyword_arguments
+
+      each_iteration(object_from_enumerator, *positional_arguments, **keyword_arguments)
+    end
+
+    # Check if the given `method_name`'s declaration accepts keyword arguments,
+    # and split up forwarded keyword arguments when any were passed to the job.
+    #: (Array[top], Symbol, top?) -> [Array[top], top?]
+    def extract_keyword_arguments(arguments, method_name, forwarded_keyword_arguments)
```
paracycle: This is usually named `normalize_arguments` and if this will be doing the kwargs to args mapping as well, then it will be a better name for it.

paracycle (commented):
I have a few stylistic comments

guides/argument-semantics.md
SomalianIvan: good point, there is nothing special about this and it should be understood implicitly, removed this

lib/job-iteration/iteration.rb
SomalianIvan: Great - I think we are aligned after this [discussion](https://github.com/Shopify/job-iteration/pull/715#discussion_r3352036242). After this backwards compatible lands, we can remove the entire proxy call in a follow up! We good to resolve @paracycle @rafaelfranca ?

test/unit/active_job_iteration_test.rb
SomalianIvan: Agreed, that coverage is missing, added 2 new tests

lib/job-iteration/iteration.rb
SomalianIvan: I refactored into `normalize_arguments` based on this suggestion https://github.com/Shopify/job-iteration/pull/715#discussion_r3352220847. I think it looks a lot better this way 👍

lib/job-iteration/iteration.rb
SomalianIvan: refactored! 👍

lib/job-iteration/iteration.rb
```diff
@@ -178,8 +178,8 @@ def interruptible_perform(*arguments)
       end
     end
 
-    def iterate_with_enumerator(enumerator, arguments)
-      arguments = arguments.dup.freeze
+    def iterate_with_enumerator(enumerator, args, kwargs)
+      args = args.dup.freeze
```
Mangara: We should dup and freeze kwargs too, right? We don't want `each_iteration` modifying those.

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +215,63 @@ def iterate_with_enumerator(enumerator, arguments)
...
+      each_iteration(object_from_enumerator, *positional_args, **keyword_args)
+    end
+
+    # Normalize Active Job kwargs for job-iteration dispatch. If the target method
+    # accepts job keyword parameters other than build_enumerator's reserved cursor:,
+    # keep kwargs separate so they can be splatted. Otherwise, append kwargs as a
+    # positional params Hash for transitional compatibility with existing jobs
+    # enqueued with keyword syntax.
+    #: (Array[top], Hash[Symbol, top], Symbol) -> [Array[top], Hash[Symbol, top]?]
+    def normalize_arguments(args, kwargs, method_name)
```
Mangara: We have to be a little careful here: these methods are getting mixed directly into the job class, so users could accidentally override them, trying to write business logic for their job. Maybe `normalize_job_iteration_arguments`?

Mangara (approved):
Nice work!

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +215,63 @@ def iterate_with_enumerator(enumerator, arguments)
...
+
+      if keyword_args.key?(:cursor)
+        raise ArgumentError, "The keyword argument `cursor` is reserved for the job iteration framework. " \
+          "Please remove `cursor` from the arguments passed to the job or rename it"
+      end
+
+      build_enumerator(*positional_args, **keyword_args, cursor: cursor_position)
+    end
+
+    def call_each_iteration(object_from_enumerator, args, kwargs)
```
liamstewart: Double-splatting nil or {} is fine so I think this can either be simplified to
```suggestion
    def call_each_iteration(object_from_enumerator, args, kwargs)
      positional_args, keyword_args = normalize_arguments(args, kwargs, :each_iteration)
      each_iteration(object_from_enumerator, *positional_args, **keyword_args)
    end
```

or inlined into the one callsite.

liamstewart (approved):

lib/job-iteration/iteration.rb
```diff
@@ -178,8 +178,9 @@ def interruptible_perform(*arguments)
       end
     end
 
-    def iterate_with_enumerator(enumerator, arguments)
-      arguments = arguments.dup.freeze
+    def iterate_with_enumerator(enumerator, args, kwargs)
+      args = args.dup.freeze
+      kwargs = kwargs.dup.freeze
```
paracycle: Small nitpick: Instead of dup'ing and freezing args and kwargs here, I think we can make `normalize_arguments` to always create a copy of them (it already does in the else case) and call freeze on the return value. Since all argument handling has to go through that method, we won't have to explicit do that anymore.

lib/job-iteration/iteration.rb
```diff
@@ -215,6 +216,60 @@ def iterate_with_enumerator(enumerator, arguments)
...
+    #: (Array[top], Hash[Symbol, top], Symbol) -> [Array[top], Hash[Symbol, top]?]
+    def normalize_arguments(args, kwargs, method_name)
+      if kwargs.empty?
+        [args, nil]
+      elsif job_has_keyword_parameters?(method_name)
+        [args, kwargs]
+      else
+        params_hash_args = args + [kwargs]
+        [params_hash_args, nil]
+      end
```
paracycle, line 251: In line with my comment above:

```suggestion
      if kwargs.empty?
        [args.dup.freeze, nil]
      elsif job_has_keyword_parameters?(method_name)
        [args.dup.freeze, kwargs.dup.freeze]
      else
        params_hash_args = args + [kwargs]
        [params_hash_args.freeze, nil]
      end
```

lib/job-iteration/iteration.rb
SomalianIvan: renamed!

lib/job-iteration/iteration.rb
SomalianIvan: Combined your suggestion to freeze kwargs and @paracycle' [feedback](https://github.com/Shopify/job-iteration/pull/715#discussion_r3357327164) to move freezing into `normalize_job_iteration_arguments`

lib/job-iteration/iteration.rb
SomalianIvan: done!

SomalianIvan:
@liamstewart I tried to apply your suggestion for simplifying `**nil` case, but unfortunately it doesn't work on ruby < 3.4, added a comment to clarify

paracycle:
> @liamstewart I tried to apply your suggestion for simplifying `**nil` case, but unfortunately it doesn't work on ruby < 3.4, added a comment to clarify

Yeah, it was added here: https://github.com/ruby/ruby/commit/5c823aa686a5549649df4af86d173bebed2418e1 in response to this ticket https://bugs.ruby-lang.org/issues/20064 so it is only available on 3.4 and 4.0.

SomalianIvan:
@paracycle what are the conventions for dropping ruby support? should we drop < 3.4 in the next major version?

paracycle:
> @paracycle what are the conventions for dropping ruby support? should we drop < 3.4 in the next major version?

At Shopify, we usually drop it when a Ruby version becomes end of life. 3.2 has just become EoL recently, and 3.3 will be next year, so we can drop 3.3 support next year. See this for reference: https://endoflife.date/ruby

Of course, this gem can do whatever it feels like is best for the codebase, but I find dropping support for 3.3 when it is supported upstream just to get support for `**nil` a little too much.

We could do `**(kwargs || {})` instead though.

SomalianIvan:
> We could do **(kwargs || {}) instead though.

Awesome!! Applied the suggestion.

paracycle (approved):
Thanks!

rafaelfranca (approved):

PR #722, opened 2026-06-09
feat: add interruption adapter for shoryuken

### LINKED ISSUE #719: Support Shoryuken
Hi, I am currently using maintenance_tasks with [Shoryuken](https://github.com/ruby-shoryuken/shoryuken). 

It appears that the job-iteration gem does not currently define an interruption_adapter for Shoryuken. Can we add one? 
I have made a PR with it, but I can't push to the repo. The code is something like this:

```rb
# frozen_string_literal: true

begin
  require "shoryuken"
rescue LoadError
  # Shoryuken is not available, no need to load the adapter
  return
end

begin
  # Lifecycle event registration (Shoryuken.configure_server / config.on) was
  # introduced in Shoryuken 2.0.2.
  gem("shoryuken", ">= 2.0.2")
rescue Gem::LoadError
  warn("job-iteration's interruption adapter for Shoryuken requires Shoryuken 2.0.2 or newer")
  return
end

module JobIteration
  module InterruptionAdapters
    module ShoryukenAdapter
      class << self
        attr_accessor :stopping

        def call
          stopping
        end
      end

      Shoryuken.configure_server do |config|
        config.on(:shutdown) do
          ShoryukenAdapter.stopping = true
        end
      end
    end

    register(:shoryuken, ShoryukenAdapter)
  end
end

```

The rest of the code is quite simple: 
1) Add shoryuken in `interruption_adapters.rb` BUNDLED_ADAPTERS array.
2) Update CHANGELOG and README
3) Add shoryuken to the Gemfile

How can I contribute? Thanks in advance

luizlevindiniz:
close: #719 
This PR adds an interruption adapter for [shoryuken](https://github.com/ruby-shoryuken/shoryuken)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,7 @@ end
 gem "sidekiq", sidekiq_version
 gem "resque", resque_version
 gem "delayed_job"
+gem "shoryuken"
 gem "connection_pool", connection_pool_version
 
 if defined?(@rails_gems_requirements) && @rails_gems_requirements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ nil
 
 ### Features
 
-nil
+- Add interruption adapter for [Shoryuken](https://github.com/ruby-shoryuken/shoryuken).
 
 ### Bug fixes
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -53,6 +53,21 @@ GEM
       tzinfo (~> 2.0, >= 2.0.5)
       uri (>= 0.13.1)
     ast (2.4.3)
+    aws-eventstream (1.4.0)
+    aws-partitions (1.1261.0)
+    aws-sdk-core (3.252.0)
+      aws-eventstream (~> 1, >= 1.3.0)
+      aws-partitions (~> 1, >= 1.992.0)
+      aws-sigv4 (~> 1.9)
+      base64
+      bigdecimal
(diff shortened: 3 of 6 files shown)


luizlevindiniz:
I have signed the CLA!

Gemfile
```diff
@@ -28,6 +28,8 @@ end
 gem "sidekiq", sidekiq_version
 gem "resque", resque_version
 gem "delayed_job"
+gem "shoryuken"
+gem "aws-sdk-sqs"
```
Mangara: We don't actually have integration tests against these gems. Do they need to be here?

Mangara (approved):
Thanks for adding this! One question, otherwise LGTM.

Gemfile
luiz-felipe-gls: Good question. 

`Aws-sdk-sqs` seems to be unnecessary here, so I have just removed it in a new commit. 

Regarding shoryuken, removing it fails `InterruptionAdaptersTest#test_loads_all_available_interruption_adapters(test/integration/interruption_adapters_test.rb)`. If you prefer, I can add shoryuken to the `adapters_to_exclude` array in this same file.

luiz-felipe-gls: We can test both approaches. In fact, I have just remembered that I have excluded aws-sdk-sqs from the gemfile in a previous commit, but it was causing some errors in the ci (https://github.com/Shopify/job-iteration/actions/runs/27232224197/job/80415276217)

PR #737, opened 2026-09-16
Constraint json on Ruby 3.1 and upgrade multi_json

vinistock:
JSON 3 has some breaking changes that affect `multi_json`, which are fixed in the newest release of `multi_json`. However, the latest `multi_json` has a >= 3.2 Ruby version requirement and this gem still supports Ruby 3.1.

I added a new Gemfile constraint to make sure that in Ruby 3.1, we don't try to use JSON 3 since we can't use the latest `multi_json`. For newer versions, I just upgraded `multi_json`, which required upgrading `mysql2` as well.

This PR fixes the CI failures on main right now.


Gemfile
```diff
@@ -68,6 +68,14 @@ else
   sorbet_version = "< 0.6.12698"
 end
 
+json_version = if ruby_version < Gem::Version.new("3.2")
+  "< 3"
+else
+  ">= 3"
+end
+
+gem "json", json_version
```
Morriar: nit:

```suggestion
gem "json", ruby_version < Gem::Version.new("3.2") ?  "< 3" : ">= 3"
```

Morriar (approved):

Mangara (approved):

PR #742, opened 2026-09-18
Ignore Active Record schema queries by notification name

selesse:
Rails edge now reads MySQL column metadata through `information_schema` instead of `SHOW FULL FIELDS`. The test helper's adapter-specific filter therefore leaves a schema query as `queries.first`, breaking the composite-primary-key assertion.

Filter `sql.active_record` events using Active Record's `"SCHEMA"` notification name, which is shared by every supported Rails version.

This unblocks https://github.com/Shopify/job-iteration/pull/740.


test/unit/active_record_batch_enumerator_test.rb
```diff
@@ -265,7 +265,7 @@ def build_enumerator(relation: Product.all, batch_size: 2, timezone: nil, column
     def track_queries(&block)
       queries = []
       query_cb = ->(*, payload) {
-        return if /SHOW FULL FIELDS FROM `\w+`/.match?(payload[:sql])
+        return if payload[:name] == "SCHEMA"
```
Mangara, line 268: Is this check compatible with all supported Rails versions?

Mangara (commented):
I don't think we need the `json` change any more after https://github.com/Shopify/job-iteration/pull/737.

test/unit/active_record_batch_enumerator_test.rb
selesse: Yep - `SCHEMA` is Active Record’s own query name across every supported version, not something we infer from the SQL. Rails 7.1–8.1 pass it explicitly, and [edge’s `query_all` defaults to it](https://github.com/rails/rails/blob/2e11aac5950afcf929aabecb6120e95bc26f3609/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb#L137-L140). The full matrix is green with this change too.

selesse:
I dropped the JSON change, I hadn't seen https://github.com/Shopify/job-iteration/pull/737. 👍

